PR #393, opened 2024-08-29
ETH-USD price fallback

### LINKED ISSUE #474: CS-BOLD-010 Low 5.13: Shutdown Can Be Triggered Twice
The shutdown function in `BorrowerOperations` is used to shut down the branch when the `TCR` falls below the `SCR`. If the branch is already shut down, the function will revert.

However, the function calls `priceFeed.fetchPrice()`, which can also shut down the branch if the price feed is failing.

This means that if the oracle fails at the same time that the branch's `TCR` falls below the `SCR`, the branch can be shut down twice. This will emit both the `ShutDown` and `ShutDownFromOracleFailure` events, which should be mutually exclusive and may not be handled correctly by off-chain infrastructure. There don't seem to be any other side effects, as the `_applyShutdown` will have no additional effect if it is called twice in the same block.

RickGriff:
Addresses the issue of a stale price during shutdown after oracle failure. For WSTETH and RETH branches, we now do our best to use a live price based on ETH-USD x canonical_rate if the primary price calculation has failed.
https://github.com/liquity/bold?tab=readme-ov-file#4---oracle-failure-and-urgent-redemptions-with-the-frozen-last-good-price

Fixes #474


contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
```diff
@@ -70,11 +83,11 @@ abstract contract MainnetPriceFeedBase is IPriceFeed, Ownable {
         return (scaledPrice, oracleIsDown);
     }
 
-    function _disableFeedAndShutDown(address _failedOracleAddr) internal returns (uint256) {
+    function _shutDownAndSwitchToLastGoodPrice(address _failedOracleAddr) internal returns (uint256) {
         // Shut down the branch
         borrowerOperations.shutdownFromOracleFailure(_failedOracleAddr);
```
bingen, line 88: It’s not related to this PR, but I wonder if we could emit the `ShutDownFromOracleFailure` here, so we don’t need to pass `_failedOracleAddr` as param to `BorrowerOperations`.

contracts/src/PriceFeeds/CompositePriceFeed.sol
```diff
@@ -4,73 +4,72 @@ pragma solidity 0.8.18;
...
+                if(ethUsdOracleDown) {
+                    // No need to shut down, since branch already is shut down
+                    return (lastGoodPrice, false);
+                } else {
+                    return (_fetchPriceETHUSDxCanonical(ethUsdPrice), false);
+                }
+            }
+
+        // Otherwise if branch is shut down and already using the lastGoodPrice, continue with it
+        if (priceSource == PriceSource.lastGoodPrice) {return (lastGoodPrice, false);}
```
bingen, line 46: Maybe we can get rid of the `if`, as there’s no other possibility, and note it in a comment.
It doesn’t  matter too much, but right now it feels like the execution flow could not enter in any of the `if` clauses and return `(0, false)` (which is not the case, fortunately).

contracts/src/PriceFeeds/CompositePriceFeed.sol
```diff
@@ -4,73 +4,72 @@ pragma solidity 0.8.18;
...
+        priceSource = PriceSource.ETHUSDxCanonical;
+        return _fetchPriceETHUSDxCanonical(_ethUsdPrice);
+    }
 
+    // Only called if the primary LST oracle has failed, branch has shut down, 
+    // and we've switched to using: ETH-USD * canonical_rate.
+    function _fetchPriceETHUSDxCanonical(uint256 _ethUsdPrice) internal returns (uint256) {
+        assert(priceSource == PriceSource.ETHUSDxCanonical);
         // Get the ETH_per_LST canonical rate directly from the LST contract
         // TODO: Should we also shutdown if the call to the canonical rate reverts, or returns 0?
```
bingen, line 62: I think so, we should fallback to `lastGoodPrice`.

contracts/src/PriceFeeds/CompositePriceFeed.sol
```diff
@@ -4,73 +4,72 @@ pragma solidity 0.8.18;
...
-
-        // Take the minimum of (market, canonical) in order to mitigate against upward market price manipulation
-        uint256 lstUsdPrice = LiquityMath._min(lstUsdMarketPrice, lstUsdCanonicalPrice);
+        uint256 lstUsdCanonicalPrice = _ethUsdPrice * lstEthRate / 1e18;
 
-        lastGoodPrice = lstUsdPrice;
+        lastGoodPrice = lstUsdCanonicalPrice;
 
-        return (lstUsdPrice, false);
+        return (lstUsdCanonicalPrice);
```
bingen, line 70: I guess we don’t parenthesis anymore.

contracts/src/PriceFeeds/CompositePriceFeed.sol
```diff
@@ -4,73 +4,72 @@ pragma solidity 0.8.18;
...
+        return _fetchPriceETHUSDxCanonical(_ethUsdPrice);
+    }
 
+    // Only called if the primary LST oracle has failed, branch has shut down, 
+    // and we've switched to using: ETH-USD * canonical_rate.
+    function _fetchPriceETHUSDxCanonical(uint256 _ethUsdPrice) internal returns (uint256) {
+        assert(priceSource == PriceSource.ETHUSDxCanonical);
         // Get the ETH_per_LST canonical rate directly from the LST contract
         // TODO: Should we also shutdown if the call to the canonical rate reverts, or returns 0?
         uint256 lstEthRate = _getCanonicalRate();
```
bingen, line 63: In the case of `wstETH` this is not LST -> ETH, but LST -> stETH. I guess we are assuming `1 stETH = 1 ETH`, but historically there have been  deviations. 
https://www.coingecko.com/en/coins/lido-staked-ether/eth
Maybe we should at least note it as a comment or in the README.

contracts/src/PriceFeeds/CompositePriceFeed.sol
```diff
@@ -4,73 +4,72 @@ pragma solidity 0.8.18;
...
 
         // Calculate the canonical LST-USD price: USD_per_LST = USD_per_ETH * ETH_per_LST
-        uint256 lstUsdCanonicalPrice = ethUsdPrice * lstEthRate / 1e18;
-
-        // Take the minimum of (market, canonical) in order to mitigate against upward market price manipulation
-        uint256 lstUsdPrice = LiquityMath._min(lstUsdMarketPrice, lstUsdCanonicalPrice);
+        uint256 lstUsdCanonicalPrice = _ethUsdPrice * lstEthRate / 1e18;
```
bingen, line 66: I wonder if it would still make sense to use the min against `lastGoodPrice`.
If the branch was shutdown, it may be due to a bug in the LST, so the canonical price may be wrong.
This way we would at least make sure we protect against the worst case of urgent redemptions not being profitable.

contracts/src/test/TestContracts/Deployment.t.sol
```diff
@@ -158,20 +156,14 @@ contract TestDeployer is MetadataDeployment {
         address ETHOracle;
         address STETHOracle;
         address RETHOracle;
-        address ETHXOracle;
-        address OSETHOracle;
```
bingen, line 162: :cry:

bingen (approved):
Looks good in general. I left some questions inline.

contracts/src/PriceFeeds/WETHPriceFeed.sol
```diff
@@ -3,35 +3,42 @@
...
+        assert(priceSource == PriceSource.primary);
+    }
+
+     function fetchPrice() public returns (uint256, bool) {
+        // If branch is live and the primary oracle setup has been working, try to use it 
+        if (priceSource == PriceSource.primary) {
+            return _fetchPrice();}
+
+        // Otherwise if branch is shut down and already using the lastGoodPrice, continue with it
+        if (priceSource == PriceSource.lastGoodPrice) {return (lastGoodPrice, false);}
```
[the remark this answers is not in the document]
RickGriff: Good point.

contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
RickGriff: Sure, makes sense

contracts/src/PriceFeeds/CompositePriceFeed.sol
RickGriff: Yes, it's the best we can do when the STETH-USD source has failed. Sure, I'll make a note in the Readme.

contracts/src/PriceFeeds/CompositePriceFeed.sol
RickGriff: Right, yes this makes sense. It could still be that `market_price < min(lastGoodPrice, canonical_price)` (depends on price movement) - but it increases the chances that we can fully clear the bad debt. Good idea!

contracts/src/PriceFeeds/WETHPriceFeed.sol
```diff
@@ -3,35 +3,42 @@
...
 
         // Check the oracle didn't already fail
-        assert(priceFeedDisabled == false);
+        assert(priceSource == PriceSource.primary);
+    }
+
+     function fetchPrice() public returns (uint256, bool) {
+        // If branch is live and the primary oracle setup has been working, try to use it 
+        if (priceSource == PriceSource.primary) {
+            return _fetchPrice();}
```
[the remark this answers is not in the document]
bingen: This is not done yet, right? No big deal, we can merge and do `forge fmt` later.

contracts/src/PriceFeeds/WETHPriceFeed.sol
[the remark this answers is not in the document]
bingen: This is not done yet, right?

contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
bingen: This is not done yet, right?

contracts/src/PriceFeeds/CompositePriceFeed.sol
bingen: This is not done yet, right?

contracts/src/PriceFeeds/CompositePriceFeed.sol
bingen: This is not done yet, right?

contracts/src/PriceFeeds/CompositePriceFeed.sol
bingen: This is not done yet, right?

contracts/src/PriceFeeds/CompositePriceFeed.sol
bingen: This is not done yet, right?

contracts/src/PriceFeeds/CompositePriceFeed.sol
bingen: This is not done yet, right?

contracts/src/PriceFeeds/CompositePriceFeed.sol
bingen: I guess we can remove the TODO?

contracts/src/PriceFeeds/RETHPriceFeed.sol
```diff
@@ -16,53 +16,67 @@ contract RETHPriceFeed is CompositePriceFeed, IRETHPriceFeed {
...
 
         // Otherwise, use the primary price calculation:
 
-        // Calculate the market LST-USD price: USD_per_LST = USD_per_ETH * ETH_per_LST
+        // Calculate the market LST-USD price: USD_per_RETH = USD_per_ETH * ETH_per_RETH
         uint256 lstUsdMarketPrice = ethUsdPrice * rEthEthPrice / 1e18;
 
-        // Calculate the canonical LST-USD price: USD_per_LST = USD_per_ETH * ETH_per_LST
+        // Calculate the canonical LST-USD price: USD_per_RETH = USD_per_ETH * ETH_per_RETH
         // TODO: Should we also shutdown if the call to the canonical rate reverts, or returns 0?
```
bingen, line 55: We are doing this above, right? So we can remove the TODO

bingen (approved):
Looks good to me!

contracts/src/PriceFeeds/RETHPriceFeed.sol
```diff
@@ -48,21 +50,37 @@ contract RETHPriceFeed is CompositePriceFeed, IRETHPriceFeed {
...
 
-        // Take the minimum of (market, canonical) in order to mitigate against upward market price manipulation.
-        // NOTE: only needed
-        uint256 lstUsdPrice = LiquityMath._min(lstUsdMarketPrice, lstUsdCanonicalPrice);
+        // If it's a redemption and canonical is within 2% of market, use the max to mitigate unwanted redemption oracle arb
+        if (_isRedemption && _within2pct(rEthUsdMarketPrice, rEthUsdCanonicalPrice)) { 
+            rEthUsdPrice = LiquityMath._max(rEthUsdMarketPrice, rEthUsdCanonicalPrice);
+        } else {
+            // Take the minimum of (market, canonical) in order to mitigate against upward market price manipulation.
+            // Assumes a deviation between market <> canonical of >2% represents a legitemate market price difference.
```
bingen, line 66: There’s a typo here, right? (legitimate)

contracts/src/PriceFeeds/RETHPriceFeed.sol
```diff
@@ -48,21 +50,37 @@ contract RETHPriceFeed is CompositePriceFeed, IRETHPriceFeed {
...
-        return (lstUsdPrice, false);
+        return (rEthUsdPrice, false);
     }
 
+    function _within2pct(uint256 _rEthUsdMarketPrice, uint256 _rEthUsdCanonicalPrice) internal pure returns (bool) {
+        // Calculate the price deviation of the oracle market price relative to the canonical price
+        uint256 max = _rEthUsdCanonicalPrice * (DECIMAL_PRECISION + _2_PERCENT) / 1e18;
+        uint256 min = _rEthUsdCanonicalPrice * (DECIMAL_PRECISION - _2_PERCENT) / 1e18;
+
+        return _rEthUsdMarketPrice >= min && _rEthUsdCanonicalPrice <= max;
```
bingen, line 80: Shouldn’t both be `_rEthUsdMarketPrice` here?

contracts/src/PriceFeeds/WETHPriceFeed.sol
```diff
@@ -10,22 +10,27 @@ contract WETHPriceFeed is MainnetPriceFeedBase {
...
         return (lastGoodPrice, false);
     }
 
-    function _fetchPricePrimary() internal override returns (uint256, bool) {
+    function fetchRedemptionPrice() external returns (uint256, bool) {
+        // Use same price for redemption as all other ops in WETH branch
+        return fetchPrice();
+    }
+
+    function _fetchPricePrimary(bool _isRedemption) internal override returns (uint256, bool) {
```
bingen, line 33: `_isRedemption` isn’t used here? Doesn’t it throw a warning?

PR #409, opened 2024-09-04
fix: Prevent arbitrary calls to leverage zappers

### LINKED ISSUE #407: [Recon] The flashloan protection is insufficient - We can operate on Troves we don't own
Anyone can call Balancer `Vault` and set the `BalancerFlashloan` as the recipient

bingen:
Closes #407

### DIFF
--- a/contracts/src/Zappers/Modules/FlashLoans/BalancerFlashLoan.sol
+++ b/contracts/src/Zappers/Modules/FlashLoans/BalancerFlashLoan.sol
@@ -46,6 +46,9 @@ contract BalancerFlashLoan is IFlashLoanRecipient, IFlashLoanProvider {
         receiver = IFlashLoanReceiver(msg.sender);
 
         vault.flashLoan(this, tokens, amounts, userData);
+
+        // Reset receiver
+        receiver = IFlashLoanReceiver(address(0));
     }
 
     function receiveFlashLoan(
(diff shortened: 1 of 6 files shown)


contracts/src/Zappers/Modules/FlashLoans/BalancerFlashLoan.sol
```diff
@@ -34,19 +29,22 @@ contract BalancerFlashLoan is IFlashLoanRecipient, IFlashLoanProvider {
...
-            userData = abi.encode(_caller, _operation, leverDownTroveParams);
+            userData = abi.encode(_operation, leverDownTroveParams);
         } else {
             revert("LZ: Wrong Operation");
         }
 
+        // This will be used by the callback below no
+        receiver = IFlashLoanReceiver(msg.sender);
+
         vault.flashLoan(this, tokens, amounts, userData);
```
danielattilasimon, line 48: Personally, I would have put the reset of `receiver` here, after the return of `vault.flashLoan()`. IMO it's more readable when the assignments are kept in the same function (easy to pair them together).

Plus, the current implementation relies on Balancer not returning successfully without having called us back. Now, in this case, it's easy to see by reading the Balancer source code that that's impossible, but in general, I prefer to minimize relying on external "promises" as much as possible.

danielattilasimon (approved):
Looks good to me! 👍 👍

PR #410, opened 2024-09-04
Add Attributes to URI

cupOJoseph:
this is what the URI _should_ look like:

```javascript
{
  "description": "....", 
  "external_url": "...", //havent set one of these yet
  "image": "base64 stuff here", 
  "name": "...",
  "attributes": 
[
    {"trait_type": "Collateral Token", "value": 0xabc123},
    {"trait_type": "Collateral Amount", "value": 5000000000000000},
    ...
 ]
}
```

json is now valid and contains useful info. Example output:
https://gist.github.com/cupOJoseph/ac67c2af41e77f2e829f9a5bd0aa9140

We should have some direct tests of the output too possibly, so I added a to-do. in the test file.

### DIFF
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -27,6 +27,8 @@ $ forge build
 $ forge test
 ```
 
+Run tests with `forge test -vvv` to see the console logs, which will show trove URI data.
+
 ### Format
 
 ```shell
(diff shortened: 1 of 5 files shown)


cupOJoseph:
looks like failing tests might be unrelated? I think this is ready for review

ColinPlatt (approved):

PR #411, opened 2024-09-04
[Recon] Zappers leverage exchange leftovers

### LINKED ISSUE #433: Zappers need to be adapted to partial repayment
In case the initial repayment amount is too high, the remainder should be sent back to user. So the zappers need to be adapted to this change:
https://github.com/liquity/bold/pull/309

### LINKED ISSUE #465: CS-BOLD-003 High 5.3: Leverage Zappers Do Not Return Swap Excess
Duplicate of #466

### LINKED ISSUE #466: Recon M-06: Zappers and swaps can leave dust
Zapper can leave dust amounts under two conditions:

- Debt to repay is less than the amount -> Dust amount, QA in most cases
- Borrowed amount is more than enough to repay the flashloan -> Will result in a significant amount of WETH being stuck

Mitigation consists of checking against any dust amount and sweeping them to the user

### LINKED ISSUE #467: CS-BOLD-004 High 5.4: Zappers Can Lose User Funds
Duplicate of #433 

The adjust, leverdown, and repay functions in the zappers can reduce the debt of a user's trove. To do
this, they transfer the specified amount to repay from the user and then trigger the repayment in the
BorrowerOperations.
However, if the specified debt repayment amount would bring a trove below MIN_DEBT, only the amount
that would bring the trove to MIN_DEBT is repaid. The rest of the funds will remain stuck in the zapper.
This is due to the following code in _adjustTrove():
```
// When the adjustment is a debt repayment, check it's a valid amount and that the caller has enough Bold
if (_troveChange.debtDecrease > 0) {
    uint256 maxRepayment = vars.trove.entireDebt > MIN_DEBT ? vars.trove.entireDebt - MIN_DEBT : 0;
    if (_troveChange.debtDecrease > maxRepayment) {
        _troveChange.debtDecrease = maxRepayment;
    }
    _requireSufficientBoldBalance(vars.boldToken, msg.sender, _troveChange.debtDecrease);
}
```

Consider the following example:
1. A user has a trove with a debt of 3000 BOLD.
2. The user calls the adjust function in the zapper with a debt repayment amount of 2000 BOLD.
3. The zapper transfers 2000 BOLD from the user to the zapper.
4. The zapper triggers the repayment in the BorrowerOperations. This will repay 1000 BOLD to
the trove, bringing the debt to 2000 BOLD (the MIN_DEBT).
5. The remaining 1000 BOLD will remain stuck in the zapper.

The worst case is the following: When there is a debt repayment to one of the lowest interest rate troves
in a branch, the repayment can be frontrun by a redemption that brings the trove just above `MIN_DEBT`.

This will result in the entire intended repayment amount being stuck in the zapper. A malicious batch
manager could intentionally reduce the interest rate of the trove's batch to make the redemption possible.
The worst-case attack would look as follows:
1. A user has a trove with a debt of 1'000'000 BOLD.
2. The user calls the adjust function in the zapper with a debt repayment amount of 500'000
BOLD.
3. The trove's batch manager frontruns the call and changes the interest rate to make the trove
low in the redemption order (or it already had a low interest rate).
4. The attacker redeems the trove until it has only 2001 BOLD debt left.

#. The zapper transfers 500'000 BOLD from the user to the zapper. 4. The zapper triggers the
repayment in the BorrowerOperations. This will repay 1 BOLD to the trove, bringing the debt to 2000
BOLD (the `MIN_DEBT`). 5. The remaining 499'999 BOLD will remain stuck in the zapper. 6. The
attacker can collect the stuck funds by backrunning using the method below

Anyone can collect the stuck funds from the zapper by creating a trove in the zapper, setting the
`receiver` role in BorrowerOperations to an address other than the zapper, then creating additional
BOLD tokens through the zapper. The newly created BOLD will go to the receiver, and the zapper will
send the stuck funds to the caller.

bingen:
Fixes #433 
Fixes #465 
Fixes #466 
Fixes #467


contracts/src/Zappers/LeverageLSTZapper.sol
```diff
@@ -55,10 +59,14 @@ contract LeverageLSTZapper is GasCompZapper, IFlashLoanReceiver, ILeverageZapper
     function openLeveragedTroveWithRawETH(OpenLeveragedTroveParams calldata _params) external payable {
         require(msg.value == ETH_GAS_COMPENSATION, "LZ: Wrong ETH");
 
+        IERC20 collTokenCached = collToken;
+
+        // Set initial balances to make sure there are not lefovers
+        _setInitialBalances(collTokenCached);
```
danielattilasimon, line 65: Couldn't we save the balances in local variables here, and send the leftovers from this function (after the return of `flashLoanProvider.makeFlashLoan`) avoiding the use of storage?

bingen: Ah, yes, good point!

danielattilasimon (approved):
Looks good!

PR #414, opened 2024-09-06
fix: effects of redemption missing from `BatchUpdated` event

danielattilasimon:
In case of redemptions, we were emitting the `BatchUpdated` event _before_ performing any redemptions within the batch, thus the coll/debt values emitted reflected only the applied interest + management fees, but none of the coll/debt reduction coming from redemptions.

In every other case, we adhere to the following event sequence:

1. `(Batched)?TroveUpdated`
2. `TroveOperation`
3. `(BatchUpdated)?`

Let's do the same for redemptions. Although — technically — a single `BatchUpdated` event per batch _would_ be enough (after the last redemption within each batch) and it'd be slightly cheaper; having one for each `BatchedTroveUpdated` is going to make it easier to write queries by being able to join event tables simply on block numbers and event indices.

### DIFF
--- a/contracts/src/TroveManager.sol
+++ b/contracts/src/TroveManager.sol
@@ -620,6 +620,19 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
             _collChangeFromOperation: -int256(_singleRedemption.collLot)
         });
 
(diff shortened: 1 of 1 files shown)


bingen (approved):
Looks good.

PR #418, opened 2024-09-12
refactor: less confusing accounting in `removeFromBatch()`

danielattilasimon:
The accounting of weighted recorded debt/fee in `removeFromBatch()` has caused some confusion in the past, see for example https://github.com/liquity/bold/pull/328#discussion_r1727155098 (the comment is about redemptions, but they used to do accounting in a similar fashion).

This PR attempts to make the intent clearer.


bingen (approved):
Looks good! The fact that we don’t need to subtract the upfrontFee to the old management fee piece is nice!

PR #419, opened 2024-09-12
Hold received SP yield when total deposited amount is low

### LINKED ISSUE #317: Discrepancy between yieldGainsOwed and sum of individual yield gains in StabilityPool
See:
https://github.com/liquity/bold/blob/main/README.md#8---discrepancy-between-yieldgainsowed-and-sum-of-individual-yield-gains-in-stabilitypool

and fix here:
https://github.com/liquity/bold/pull/261

bingen:
We track those yields in a new state var `yieldGainsPending`, and we assimilate them to `yieldGainsOwed` once the `totalBoldDeposits` reaches 1 BOLD (1e18).

Closes #317

It also prevents losses of yield when the SP is empty.

### DIFF
--- a/contracts/src/ActivePool.sol
+++ b/contracts/src/ActivePool.sol
@@ -258,9 +258,11 @@ contract ActivePool is IActivePool {
             uint256 remainderToLPs = mintedAmount - spYield;
 
             _boldToken.mint(address(interestRouter), remainderToLPs);
-            _boldToken.mint(address(stabilityPool), spYield);
 
-            stabilityPool.triggerBoldRewards(spYield);
+            if (spYield > 0) {
+                _boldToken.mint(address(stabilityPool), spYield);
+                stabilityPool.triggerBoldRewards(spYield);
+            }
         }
 
         lastAggUpdateTime = block.timestamp;
--- a/contracts/src/StabilityPool.sol
+++ b/contracts/src/StabilityPool.sol
@@ -146,6 +146,9 @@ contract StabilityPool is LiquityBase, IStabilityPool, IStabilityPoolEvents {
     // TODO: from the contract's perspective, this is a write-only variable. It is only ever read in tests, so it would
     // be better to keep it outside the core contract.
     uint256 internal yieldGainsOwed;
+    // Total remaining Bold yield gains (from Trove interest mints) held by SP, not yet paid out to depositors,
+    // and not accounted for because they were received when the total deposits were too small
+    uint256 internal yieldGainsPending;
 
     // --- Data structures ---
 
@@ -225,6 +228,10 @@ contract StabilityPool is LiquityBase, IStabilityPool, IStabilityPoolEvents {
 
     function getYieldGainsOwed() external view override returns (uint256) {
         return yieldGainsOwed;
+    }
+
+    function getYieldGainsPending() external view override returns (uint256) {
+        return yieldGainsPending;
     }
 
     // --- External Depositor Functions ---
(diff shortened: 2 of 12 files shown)


contracts/src/StabilityPool.sol
```diff
@@ -263,10 +270,16 @@ contract StabilityPool is LiquityBase, IStabilityPool, IStabilityPoolEvents {
...
+        uint256 totalBoldDepositsCached = _updateTotalBoldDeposits(_topUp + keptYieldGain, 0);
         _decreaseYieldGainsOwed(currentYieldGain);
         _sendBoldtoDepositor(msg.sender, yieldGainToSend);
         _sendCollGainToDepositor(collToSend);
+
+        // If there were pending yields and with the new deposit we are reaching the threshold, let’s move the yield to owed
+        uint256 yieldGainsPendingCached = yieldGainsPending;
+        if (yieldGainsPendingCached > 0 && totalBoldDepositsCached >= DECIMAL_PRECISION) {
+            _updateYieldRewardsSum(yieldGainsPendingCached, totalBoldDepositsCached);
+        }
```
danielattilasimon, line 282: We must do the same in `withdrawFromSP()`. Reason: a depositor could have enough pending yield gains to put total deposits over the threshold of `DECIMAL_PRECISION`. Remember that yield gains are deposited if they're not claimed.

contracts/src/StabilityPool.sol
```diff
@@ -358,20 +377,27 @@ contract StabilityPool is LiquityBase, IStabilityPool, IStabilityPoolEvents {
...
         uint256 totalBoldDepositsCached = totalBoldDeposits; // cached to save an SLOAD
-        /*
-        * When total deposits is 0, B is not updated. In this case, the BOLD issued can not be obtained by later
-        * depositors - it is missed out on, and remains in the balance of the SP.
-        *
-        */
-        if (totalBoldDepositsCached == 0 || _boldYield == 0) {
+
+        // When total deposits is very small, B is not updated. In this case, the BOLD issued can not be obtained by later
+        // depositors - it is missed out on, and remains in the balance of the SP.
```
danielattilasimon, line 385: Comment is outdated now.

contracts/src/ActivePool.sol
```diff
@@ -258,9 +258,11 @@ contract ActivePool is IActivePool {
             uint256 remainderToLPs = mintedAmount - spYield;
 
             _boldToken.mint(address(interestRouter), remainderToLPs);
```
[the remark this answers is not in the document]
bingen: Actually I did that, but then removed it.
Maybe I’m missing something, but we are already checking that `mintedAmount` is not zero, so if `spYield` is zero, then `remainderToLPs` shouldn’t be zero.

contracts/src/StabilityPool.sol
bingen: Ah, right!

contracts/src/ActivePool.sol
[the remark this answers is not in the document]
danielattilasimon: Ah, yes, I forgot about the outer condition `mintedAmount > 0`. So the only way `spYield` can be zero is if the minted amount is absolutely _tiny_.

contracts/src/StabilityPool.sol
bingen: Ah, I see you already fixed it, thanks!

contracts/src/test/TestContracts/InvariantsTestHandler.t.sol
```diff
@@ -1529,24 +1540,35 @@ contract InvariantsTestHandler is BaseHandler, BaseMultiCollateralTest {
...
+        v.boldYield = v.c.stabilityPool.getDepositorYieldGainWithPending(msg.sender);
+        v.ethGain = v.c.stabilityPool.getDepositorCollGain(msg.sender);
+        v.ethStash = v.c.stabilityPool.stashedColl(msg.sender);
+        v.ethClaimed = claim ? v.ethStash + v.ethGain : 0;
+        v.boldClaimed = claim ? v.boldYield : 0;
+        uint256 totalBoldDepositsBefore = v.c.stabilityPool.getTotalBoldDeposits();
+        uint256 totalBoldDepositsAfter = amount + totalBoldDepositsBefore;
+        if (totalBoldDepositsBefore < DECIMAL_PRECISION && totalBoldDepositsAfter >= DECIMAL_PRECISION) {
+            v.depositorPendingYield =
+                (v.pendingYield + SP_YIELD_SPLIT * v.pendingInterest / 1e18) * amount / totalBoldDepositsAfter;
```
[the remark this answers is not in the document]
bingen: Thanks for fixing this too!

contracts/src/StabilityPool.sol
```diff
@@ -584,11 +612,11 @@ contract StabilityPool is LiquityBase, IStabilityPool, IStabilityPoolEvents {
 
         Snapshots memory snapshots = depositSnapshots[_depositor];
 
-        uint256 pendingSPYield = activePool.calcPendingSPYield();
+        uint256 pendingSPYield = activePool.calcPendingSPYield() + yieldGainsPending;
```
RickGriff, line 615: When there are pending yield gains, how are they awarded to depositors?

This function seems to calculate as if all pending gains were immediately awarded pro-rata, including pending yield gains.

But in reality I guess some pending yield gains might not be, right (depending on the SP size). We don't really know when they'll be rewarded. Still, it's fine as an estimate IMO.

Also, I guess and IIRC if there are pending yield gains and a depositor makes a large enough deposit, he receives all those gains, right?

RickGriff (commented):
Finally reviewing this - really nice, looks good!

PR #421, opened 2024-09-12
Contracts deployment script: fix capitalization

bpierre:


danielattilasimon (approved):
Looks good!

We could also just change the definition of `collateralContracts`, but we would probably need to adjust `deployment-artifacts-to-app-env.ts` accordingly.

PR #426, opened 2024-09-16
fix: Use last zombie trove first in a redemption sequence

### LINKED ISSUE #425: Bypass redemption mechanism with min interest rate by splitting a trove into multive zombie ones
Brought up by @Javalasers on Discord:
## Description

> There's an attack vector where the user can:
> * Create a trove with 2000 BOLD of debt and ~0%~ `MIN_ANNUAL_INTEREST_RATE` (0.5%?) interest.
> * Redeem 1 BOLD worth of collateral from the protocol (let's say they redeem from their own trove).
> * They now have a trove paying 0% interest but with 1999 BOLD of debt.
> 
> So if I want to mint 20,000 BOLD against rETH collateral, it seems I could do the operation above 10 times, and hold that position while paying 0 protocol interest, but being immune from redemptions (nullifying the usual incentive that I have to set a nonzero interest rate). I'm sure I'm missing something; just started reading about this in earnest...

## Impact:

If this attack is profitable, it may break the whole equilibrium and game theory of the user set interest rate.

Some math should be done, and it heavily depends on gas price, but some considerations:
- The interest wouldn’t be zero, as this is not possible, so eventually the trove would become redeemable again if too close to `MIN_DEBT`. So if the aim is to keep it 1 year, at min interest rate of 0.5%, the zombie troves should have 1,990 BOLD debt.
- To set it up `~2n` txs are needed, where `n ~= total debt / MIN_DEBT`
- There’s an additional `(n-1) * ETH_GAS_COMPENSATION` ETH that needs to be locked up.
- Also, the management of the trove needs to be considered, so:
  - Either they all have a very high CR -> not capital efficient, it’s kind of a hidden cost
  - Or then `2n` times more gas have to be spent to adjust troves every time if the attacker is trying to have a low CR and actively manage them (for each trove there’s 1 adjustment that brings it up above `MIN_DEBT` and then again a redemption).

## Potential solutions

1. Keep a pointer to the last redemeed zombie trove, and use that one first when the next redemption happens (h/t @danielattilasimon ). The attack would still work with redistribution, but that’s way more unlikely and harder to perform (with an extra cost of 5% of the SP size, a huge trove to be liquidated could be opened and then go for the redistributions, but it’s hard, expensive, can’t be done in 1 tx, it depends on ETH fluctuations, on the global TCR...).
2. When a trove gets redeemed and ends up with non zero debt, automatically set it to the max interest rate (h/t @cvalkan )

bingen:
Closes #425

Note: with this implementation, it can happen that a trove belonging to a batch, has both entire and recorded debt, due to the batch being updated (even the function `redeemCollateral` does that as a preliminary step) but without a direct action on the trove, and the trove is still zombie and holds the pointer to the last zombie trove, so it gets redeemed first in the list.

This seems like quite an edge case, and the user can easily prevent it by taking care of the trove before.

### DIFF
--- a/contracts/src/Interfaces/ITroveManager.sol
+++ b/contracts/src/Interfaces/ITroveManager.sol
@@ -18,7 +18,7 @@ interface ITroveManager is ILiquityBase {
         active,
         closedByOwner,
         closedByLiquidation,
-        unredeemable
+        zombie
     }
 
     function shutdownTime() external view returns (uint256);
@@ -52,6 +52,8 @@ interface ITroveManager is ILiquityBase {
     function getTroveFromTroveIdsArray(uint256 _index) external view returns (uint256);
 
     function getCurrentICR(uint256 _troveId, uint256 _price) external view returns (uint256);
+
+    function lastZombieTroveId() external view returns (uint256);
 
     function batchLiquidateTroves(uint256[] calldata _troveArray) external;
 
@@ -88,7 +90,7 @@ interface ITroveManager is ILiquityBase {
         uint256 _batchDebt
     ) external;
 
-    // Called from `adjustUnredeemableTrove()`
+    // Called from `adjustZombieTrove()`
     function setTroveStatusToActive(uint256 _troveId) external;
 
     function onAdjustTroveInterestRate(
--- a/contracts/src/TroveManager.sol
+++ b/contracts/src/TroveManager.sol
@@ -111,6 +111,8 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
     // Array of all batch managers - used to fetch them off-chain
     address[] public batchIds;
 
+    uint256 public lastZombieTroveId;
+
     // Error trackers for the trove redistribution calculation
     uint256 internal lastCollError_Redistribution;
     uint256 internal lastBoldDebtError_Redistribution;
@@ -149,6 +151,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         uint256 oldWeightedRecordedDebt;
         uint256 newWeightedRecordedDebt;
         uint256 newStake;
+        bool isZombieTrove;
         LatestTroveData trove;
         LatestBatchData batch;
     }
@@ -452,7 +455,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
     }
 
     function _isLiquidatableStatus(Status _status) internal pure returns (bool) {
-        return _status == Status.active || _status == Status.unredeemable;
+        return _status == Status.active || _status == Status.zombie;
     }
 
     function _batchLiquidateTroves(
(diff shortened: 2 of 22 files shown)


contracts/src/Interfaces/ITroveManager.sol
```diff
@@ -53,6 +53,8 @@ interface ITroveManager is ILiquityBase {
 
     function getCurrentICR(uint256 _troveId, uint256 _price) external view returns (uint256);
 
+    function lastZombieTroveId() external view returns (uint256);
```
RickGriff, line 56: "last" here refers to the last Trove that was redeemed to below MIN_DEBT right?

contracts/src/TroveManager.sol
```diff
@@ -663,14 +666,26 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
...
+                if (isTroveInBatch) {
+                    sortedTroves.removeFromBatch(_singleRedemption.troveId);
+                } else {
+                    sortedTroves.remove(_singleRedemption.troveId);
+                }
+                // If it’s a partial redemption, let’s store a pointer to it so it’s used first in the next one
+                if (newDebt > 0) {
+                    lastZombieTroveId = _singleRedemption.troveId;
+                }
+            } else if (newDebt == 0) {
```
RickGriff, line 682: OK, so `newDebt == 0` implies _this_ Trove was fully redeemed, therefore we unzombify it

contracts/src/TroveManager.sol
```diff
@@ -730,15 +745,27 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         uint256 remainingBold = _boldamount;
 
         SingleRedemptionValues memory singleRedemption;
-        singleRedemption.troveId = sortedTrovesCached.getLast();
+        // Let’s check if there’s a pending zombie trove from previous redemption
+        if (lastZombieTroveId != 0) {
```
RickGriff, line 749: We can never have a valid troveId == 0, right?

contracts/src/TroveManager.sol
```diff
@@ -769,6 +796,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
 
             remainingBold -= singleRedemption.boldLot;
             singleRedemption.troveId = nextUserToCheck;
+            singleRedemption.isZombieTrove = false;
```
RickGriff, line 799: We only need to do this if it was a `zombieTrove`. Though seems fine to just do it here at the end.

contracts/src/TroveManager.sol
```diff
@@ -1450,6 +1483,8 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         } else {
             if (trove.status == Status.active) {
                 sortedTroves.remove(_troveId);
+            } else if (trove.status == Status.zombie && lastZombieTroveId == _troveId) {
+                lastZombieTroveId = 0;
```
RickGriff, line 1487: Do we need this logic twice? Could we do it once at the end?

RickGriff (commented):
Looks great! Nice solution.

contracts/src/TroveManager.sol
bingen: I wouldn’t say “unzombified”. It’s still zombie, but it doesn’t hold the pointer to be the first one in the next redemption sequence anymore.

PR #428, opened 2024-09-20
fix: Allow owner to set any interest rate even if trove has delegate

### LINKED ISSUE #378: [Dedaub L5] Trove’s interest rate delegate’s restrictions affect the owner
The `BorrowerOperations::_requireInterestRateInDelegateRange` function, which is called only by `adjustTroveInterestRate`, checks if the trove’s owner has set an interest rate delegate and if this is the case requires the new interest rate to be between the min and max values set for this delegate. However, the function does not distinguish the case in which the caller is the owner and thus should be able to bypass the restrictions that exist for the delegate.

`BorrowerOperations::_requireInterestRateInDelegateRange`:

```
function _requireInterestRateInDelegateRange(
  uint256 _troveId, uint256 _annualInterestRate
) internal view {
  InterestIndividualDelegate memory individualDelegate =
    interestIndividualDelegateOf[_troveId];
  // Dedaub: the condition does not take into account the msg.sender
  if (individualDelegate.account != address(0)) {
    _requireInterestRateInRange(
      _annualInterestRate,
      individualDelegate.minInterestRate,
      individualDelegate.maxInterestRate
    );
  }
}
```

The condition should be changed to `individualDelegate.account != address(0)  && msg.sender != owner` or `individualDelegate.account == msg.sender`.

bingen:
Closes #378

### DIFF
--- a/contracts/src/BorrowerOperations.sol
+++ b/contracts/src/BorrowerOperations.sol
@@ -520,7 +520,7 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
         _requireIsNotInBatch(_troveId);
         address owner = troveNFT.ownerOf(_troveId);
         _requireSenderIsOwnerOrInterestManager(_troveId, owner);
-        _requireInterestRateInDelegateRange(_troveId, _newAnnualInterestRate);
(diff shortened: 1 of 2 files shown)


RickGriff (approved):
Looks good!

PR #443, opened 2024-10-02
test: fix inconsistencies in the definition of premature adjustment

danielattilasimon:
There were some inconsistencies in what is considered a premature adjustment between:
- HintHelpers,
- invariant tests,
- the actual implementation.

This resulted in occasional invariant test failures (`Should have failed as new ICR < MCR`). Let's harmonize everything with the actual implementation.


bingen (approved):
Looks good

PR #453, opened 2024-10-04
fix: avoid overflow in batched Trove's `weightedRecordedDebt`

danielattilasimon:
The batch's `weightedRecordedDebt` is already a product of 2 big numbers (batch total debt * annual interest rate), so it's best to avoid multiplying it by a 3rd one.


bingen (approved):
Looks good to me.

PR #455, opened 2024-10-07
Prevent shares to debt ratio in batches to be too low

bingen:
Fixes issues 42 from @GalloDaSballo audit and `CS-BOLD-001` from ChainSecurity.

There are 2 calls to `updatedBatchShares` that deserve special attention:

1. From redistributions:
If trove redistribution is bigger than the amount being redeemed, it would revert. It would even affect other branches. Unpredictable and bad, as it breaks the foundation of the system. There are 2 alternatives:
- Skip the trove. Adding complexity. Besides by definition that’s not good (could someone use it to game the system? Seems unlikely, but it opens the door), it may lead to DoS: if the batch is huge, and the pending redemption is low, it may happen that we have to skip all troves in that batch and therefore running out of gas. Seems super unlikely, as we need a huge batch, full of troves pending redistribution gains (which never happened in v1 and likely will never happen in v2), and that has a bad debt/shares ratio (unlikely too). But not nice to have anyway.
- [Chosen one] Skip the check for redemptions. The problem is that some troves may get less shares than they should coming from redistributions. Again, unlikely, hard to exploit on purpose. Besides the max debt/shares ratio still has a lot of margin (1e9), so the effect would be negligible. I don’t think it would be feasible in practice to lower that amount really low only by applying redistributions, and even then, only the debt coming from redistributions after a redemption would benefit from that. 

2. From `onApplyTroveInterest`. More or less the same happens here: it would revert if it has redistribution gains of any size. We can skip the check or revert. Here I would revert normally, to avoid that trove owner tries to benefit from it. The downsize is that the rest of the system cannot extract the pending interest from that trove permissionlessly, but I don’t think that’s a big deal, as the owner can not do much with that trove until it leaves the blocked batch.

### DIFF
--- a/contracts/src/Dependencies/Constants.sol
+++ b/contracts/src/Dependencies/Constants.sol
@@ -28,6 +28,11 @@ uint128 constant MIN_INTEREST_RATE_CHANGE_PERIOD = 1 seconds; // prevents more t
 
 uint256 constant REDEMPTION_FEE_FLOOR = _1pct / 2; // 0.5%
 
+// For the shares ratio to decrease 1e9 (1e18/MIN_BATCH_SHARES_RATIO),
+// at an average annual rate of 10%, it would take more than 217 years (log(1e9)/log(1.1))
+// at an average annual rate of 50%, it would take more than 51 years (log(1e9)/log(1.5))
+uint256 constant MIN_BATCH_SHARES_RATIO = 1e9;
+
 // Half-life of 12h. 12h = 720 min
 // (1/2) = d^720 => d = (1/2)^(1/720)
 uint256 constant REDEMPTION_MINUTE_DECAY_FACTOR = 999037758833783000;
--- a/contracts/src/Dependencies/Constants.sol
+++ b/contracts/src/Dependencies/Constants.sol
@@ -31,6 +31,8 @@ uint256 constant REDEMPTION_FEE_FLOOR = _1pct / 2; // 0.5%
 // For the debt / shares ratio to increase by a factor 1e9
 // at a average annual debt increase (compounded interest + fees) of 10%, it would take more than 217 years (log(1e9)/log(1.1))
 // at a average annual debt increase (compounded interest + fees) of 50%, it would take more than 51 years (log(1e9)/log(1.5))
+// The increase pace could be forced to be higher through an inflation attack,
+// but precisely the fact that we have this max value now prevents the attack
 uint256 constant MAX_BATCH_SHARES_RATIO = 1e9;
 
 // Half-life of 12h. 12h = 720 min
--- a/contracts/src/Dependencies/Constants.sol
+++ b/contracts/src/Dependencies/Constants.sol
@@ -28,10 +28,10 @@ uint128 constant MIN_INTEREST_RATE_CHANGE_PERIOD = 1 seconds; // prevents more t
 
 uint256 constant REDEMPTION_FEE_FLOOR = _1pct / 2; // 0.5%
 
-// For the shares ratio to decrease 1e9 (1e18/MIN_BATCH_SHARES_RATIO),
-// at an average annual rate of 10%, it would take more than 217 years (log(1e9)/log(1.1))
-// at an average annual rate of 50%, it would take more than 51 years (log(1e9)/log(1.5))
-uint256 constant MIN_BATCH_SHARES_RATIO = 1e9;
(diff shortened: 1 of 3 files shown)


contracts/src/TroveManager.sol
```diff
@@ -1784,6 +1785,12 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
                 if (_batchDebt == 0) {
                     batchDebtSharesDelta = debtIncrease;
                 } else {
+                    // For the shares ratio to decrease 1e9 (1e18/MIN_BATCH_SHARES_RATIO),
+                    // at an average annual rate of 10%, it would take more than 217 years (log(1e9)/log(1.1))
+                    // at an average annual rate of 50%, it would take more than 51 years (log(1e9)/log(1.5))
```
danielattilasimon, line 1790: This math is assuming no compounding, which is the best case scenario (longest time to reach the limit).

If the batch is being touched frequently, it'll be closer to:
```
t_1e9 = log(1e9) / r
```

where `r` is the annual interest rate, i.e. `0.1` & `0.5` in the examples above.

bingen: I guess my wording was ambiguous, but I was not assuming it was without compounding. I was assuming those “global average annual rates” would include not only compounding, but also fees.
A “global average automatic annual debt increase” (I don’t know how to call it :sweat_smile: ) of 10% sustained over the years sounds pretty high to me, let alone 50%.

contracts/src/TroveManager.sol
```diff
@@ -1784,6 +1790,9 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
                 if (_batchDebt == 0) {
                     batchDebtSharesDelta = debtIncrease;
```
RickGriff, line 1791: So in this case, we're assuming the debt increase can't increase the shares ratio beyond the max right?

I guess that should hold since even if a batch's ratio got inflated, when the last Trove exited, the batch's shares should have been reset to 0.

contracts/src/TroveManager.sol
```diff
@@ -1756,12 +1760,14 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
...
 
+    // This function will revert if there’s a total debt increase and the ratio debt / shares has exceeded the max
     function _updateBatchShares(
         uint256 _troveId,
         address _batchAddress,
         TroveChange memory _troveChange,
         uint256 _batchColl, // without trove change
-        uint256 _batchDebt // entire (with interest, batch fee), but without trove change, nor upfront fee nor redist
+        uint256 _batchDebt, // entire (with interest, batch fee), but without trove change, nor upfront fee nor redist
+        bool _checkBatchSharesRatio
```
RickGriff, line 1770: Ah, so this is not related to whether there's a net debt increase or not - just dictates whether we do the check on the resulting ratio inside the func call

contracts/src/Dependencies/Constants.sol
```diff
@@ -28,6 +28,11 @@ uint128 constant MIN_INTEREST_RATE_CHANGE_PERIOD = 1 seconds; // prevents more t
 
 uint256 constant REDEMPTION_FEE_FLOOR = _1pct / 2; // 0.5%
 
+// For the debt / shares ratio to increase by a factor 1e9
```
RickGriff, line 31: Yes this is the "natural" way. Though an inflation attack (as outlined by CS) would make it faster. But maybe we just document that in the docs rather than here.

RickGriff (approved):
Awesome, looks good!

contracts/src/TroveManager.sol
bingen: If the batch debt is zero, then the ratio debt / shares would either be zero or undefined. In the most common case, shares would also be zero, and after the update the ratio would be 1 (1e18).
Maybe we should explore if due to rounding, a batch that gets empty and then filled again, can have non zero shares and what are the implications, but here the ratio would deviate in the opposite direction, as shares would be bigger than debt.

contracts/src/TroveManager.sol
bingen: Yes, I’ll add a comment to clarify. I can change the name if you have a better suggestion.

contracts/src/Dependencies/Constants.sol
bingen: Good point. I added a small note there too, it doesn’t hurt.

PR #456, opened 2024-10-08
fix: zombie Trove pointer reset despite no new zombie

danielattilasimon:
Normally, `lastZombieTroveId` is first in line to get redeemed, however, it will be skipped if its ICR is less than 100%. Previously, we would have reset `lastZombieTroveId` to 0 if a subsequent redemption _doesn't_ result in a Trove having `0 < debt < MIN_DEBT`. However, there's no reason to reset `lastZombieTroveId` in this case, even if the chances of it escaping liquidation until its ICR recovers to >100% are slim.

### DIFF
--- a/contracts/src/TroveManager.sol
+++ b/contracts/src/TroveManager.sol
@@ -683,9 +683,6 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
                 // Reset last zombie trove pointer if the previous one was fully redeemed now
                 lastZombieTroveId = 0;
             }
-        } else {
-            // Reset last zombie trove pointer if the previous one ended up above min debt
-            lastZombieTroveId = 0;
         }
     }
 
--- a/contracts/src/TroveManager.sol
+++ b/contracts/src/TroveManager.sol
@@ -684,6 +684,9 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
                 lastZombieTroveId = 0;
             }
         }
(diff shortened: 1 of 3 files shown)


contracts/src/TroveManager.sol
```diff
@@ -683,9 +683,6 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
                 // Reset last zombie trove pointer if the previous one was fully redeemed now
                 lastZombieTroveId = 0;
             }
-        } else {
-            // Reset last zombie trove pointer if the previous one ended up above min debt
-            lastZombieTroveId = 0;
```
bingen, line 688: I guess there may be an edge case, where a zombie trove, with ICR > 100% (so not skipped), gets redistribution gains and a redemption, but the total amount ends up being above `MIN_DEBT`. Then it shouldn’t be zombie anymore, but I don’t think it’s worth the extra complexity as:
- It’s an extreme edge case
- The owner can manually resuscitate the trove to get rid of the pointer.

bingen (approved):
Good catch!

contracts/src/TroveManager.sol
danielattilasimon: Good thinking! I also don't think it needs special handling, but I will make a note in the source code.

danielattilasimon: Especially since proper resuscitation would need hints, which we lack in redemption.

PR #457, opened 2024-10-08
CollateralRegistry immutables mapping comment

ColinPlatt:


contracts/src/CollateralRegistry.sol
```diff
@@ -14,9 +14,6 @@ import "./Interfaces/ICollateralRegistry.sol";
 // import "forge-std/console2.sol";
 
 contract CollateralRegistry is ICollateralRegistry {
-    // mapping from Collateral token address to the corresponding TroveManagers
-    //mapping(address => address) troveManagers;
-    // See: https://github.com/ethereum/solidity/issues/12587
```
bingen, line 19: I wouldn’t really remove these comments, as that was the initial intent, and the rest was a workaround to be able to make them immutable.
At least I would leave the last one, that github link, to honour the source where we mostly copied it from.

ColinPlatt: left the last one in there as requested.

PR #507, opened 2024-10-09
fix: Use msg.sender instead of user-provided param in Zappers

### LINKED ISSUE #475: CS-BOLD-011 Low 5.14: User-provided transferFrom Source Address
In UniV3Exchange and CurveExchange, the `swapFromBold` and `SwapToBold` function take a `_zapper` argument. This is used in a `transferFrom` call.

```
boldTokenCached.transferFrom(_zapper, address(this), _boldAmount);
```

As the `_zapper` address is provided by the user, this can be used to pull tokens from any address that
has given approval to the Exchange contract. For example, the zappers give unlimited approval. If a
misinformed user gives approval to the Exchange contract accidentally, they could also be drained.
The zappers are not intended to ever have a balance, so the impact is limited. However, it is considered
bad practice to use `transferFrom` with a user-provided address.

The `msg.sender` address could be used instead.

bingen:
Fixes #475

### DIFF
--- a/contracts/src/Zappers/Interfaces/IExchange.sol
+++ b/contracts/src/Zappers/Interfaces/IExchange.sol
@@ -7,7 +7,7 @@ interface IExchange {
     function getBoldAmountToSwap(uint256 _boldAmount, uint256 _maxBoldAmount, uint256 _minCollAmount)
         external /* view */
         returns (uint256);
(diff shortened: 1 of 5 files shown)


danielattilasimon (approved):
Looks good to me! What was the idea behind parameterizing this in the first place?

PR #508, opened 2024-10-09
feat: Add functions to close from collateral to zappers

### LINKED ISSUE #459: Zappers: close a trove by repaying from the collateral

bingen:
Closes #459

### DIFF
--- a/contracts/src/Zappers/GasCompZapper.sol
+++ b/contracts/src/Zappers/GasCompZapper.sol
@@ -41,9 +41,8 @@ contract GasCompZapper is AddRemoveManagers, LeftoversSweep, IFlashLoanReceiver,
         flashLoanProvider = _flashLoanProvider;
         exchange = _exchange;
 
-        // Approve Coll and Bold to exchange module
+        // Approve Coll to exchange module (for closeTroveFromCollateral)
         collToken.approve(address(_exchange), type(uint256).max);
-        boldToken.approve(address(_exchange), type(uint256).max);
     }
 
     struct OpenTroveVars {
--- a/contracts/src/Zappers/GasCompZapper.sol
+++ b/contracts/src/Zappers/GasCompZapper.sol
@@ -269,7 +269,7 @@ contract GasCompZapper is AddRemoveManagers, LeftoversSweep, IFlashLoanReceiver,
 
         // Set initial balances to make sure there are not lefovers
         InitialBalances memory initialBalances;
-        _setInitialBalances(collToken, boldToken, initialBalances);
+        _setInitialBalancesAndReceiver(collToken, boldToken, initialBalances, receiver);
 
         // Flash loan coll
         flashLoanProvider.makeFlashLoan(
--- a/contracts/src/Zappers/GasCompZapper.sol
+++ b/contracts/src/Zappers/GasCompZapper.sol
@@ -10,10 +10,14 @@ import "../Interfaces/IWETH.sol";
 import "../Dependencies/AddRemoveManagers.sol";
 import "./LeftoversSweep.sol";
 import "../Dependencies/Constants.sol";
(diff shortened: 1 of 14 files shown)


contracts/src/Zappers/Interfaces/IZapper.sol
```diff
@@ -0,0 +1,37 @@
...
+        uint256 maxUpfrontFee;
+        address addManager;
+        address removeManager;
+        address receiver;
+    }
+
+    struct CloseTroveParams {
+        uint256 troveId;
+        uint256 flashLoanAmount;
+        uint256 minBoldAmount;
```
danielattilasimon, line 26: Is this used for anything? Looks unused to me, but I could be missing something.

contracts/src/Zappers/GasCompZapper.sol
```diff
@@ -261,5 +262,67 @@ contract GasCompZapper is AddRemoveManagers, LeftoversSweep {
...
+        //uint256 receivedBoldAmount = exchange.swapToBold(_effectiveFlashLoanAmount, trove.entireDebt);
+        //require(receivedBoldAmount >= trove.entireDebt, "GCZ: Not enough BOLD obtained to repay");
+
+        borrowerOperations.closeTrove(_params.troveId);
+
+        // Send coll back to return flash loan
+        collToken.safeTransfer(address(flashLoanProvider), _params.flashLoanAmount);
+
+        // Send coll left
+        collToken.safeTransfer(_params.receiver, trove.entireColl - _params.flashLoanAmount);
```
danielattilasimon, line 305: Hmm, there's a receiver parameter, but leftovers are still swept to `msg.sender`, is that intended?

danielattilasimon: Also, should we add an early check for `trove.entireColl >= _params.flashLoanAmount` to avoid panicking so late?

contracts/src/Zappers/GasCompZapper.sol
```diff
@@ -10,39 +10,40 @@ import "../Interfaces/IWETH.sol";
...
-        uint256 maxUpfrontFee;
-        address addManager;
-        address removeManager;
-        address receiver;
+        flashLoanProvider = _flashLoanProvider;
+        exchange = _exchange;
+
+        // Approve Coll and Bold to exchange module
+        collToken.approve(address(_exchange), type(uint256).max);
+        boldToken.approve(address(_exchange), type(uint256).max);
```
danielattilasimon, line 46: Why do we need a BOLD token approval? I thought we only swap coll => BOLD.

contracts/src/Zappers/Interfaces/IZapper.sol
bingen: Good catch, I replaced it by `trove.entireDebt` and forgot to remove this. Thanks!

contracts/src/Zappers/GasCompZapper.sol
bingen: I considered that, but in this case, being the reason a user provided param, I think it’s “user fault” (probably more the frontend but still :man_shrugging: ), so no big deal if some more gas is spent. Otherwise, we are making “good users” spend more unnecessary gas with an extra check.
Or even without the bad/good judgement: statistically how many times the happy path would happen vs the one where that param is wrong? I guess it wouldn’t pay off.

bingen: > Hmm, there's a receiver parameter, but leftovers are still swept to msg.sender, is that intended?

In the generic case, it’s intended, as it’s `msg.sender` who sends the original funds, but I guess it doesn’t make sense in this case, as funds are not being sent.

danielattilasimon (approved):
Looks great!

PR #513, opened 2024-10-10
Zappers minor fixes

### LINKED ISSUE #479: CS-BOLD-024 Informational 7.3: CEI Pattern Violated in Adjust Trove
The function `WETHZapper._adjustTrovePost` sends tokens after adjusting a trove. It first sends ETH via a call to the receiver and then sends BOLD tokens.

```
function _adjustTrovePost(
...
) internal {
// WETH -> ETH
if (!_isCollIncrease) {
WETH.withdraw(_collChange);
(bool success,) = _receiver.call{value: _collChange}("");
}
require(success, "WZ: Sending ETH failed");
}
// Send Bold
if (_isDebtIncrease) {
boldToken.transfer(_receiver, _boldChange);
}
}
```

The BOLD token is not reentrant, but the recipient could reenter the contract if it receives Ether.

Therefore, it is considered best practice to first send the bold tokens, and call the recipient last, following the Checks-effects-interactions (CEI) pattern.

### LINKED ISSUE #484: CS-BOLD-028 Informational 7.8:  Inconsistent Input Validation by Zappers
Liquity uses separate Zapper contracts for WETH collateral and other LST collateral tokens. The contracts are inconsistent.
The WETHZapper enforces, in _adjustTrovePre(), that debt must decrease by a positive amount:
```
require(!_isDebtIncrease || _boldChange > 0, "WZ: Increase bold amount should not be zero");
```
However, `GasCompZapper._adjustTrovePre()` is missing that check, allowing calls with `debtIncrease` set to true and `boldChange` equal to zero:

```
if (_isCollIncrease || (!_isDebtIncrease && _boldChange > 0)) {
    _requireSenderIsOwnerOrAddManager(_troveId, owner);
}
```

If no add manager is assigned (set to address 0), anyone can call this function, but it has no effect on `BorrowerOperations.adjustTrove()`, as `_adjustTrove()` debt increases only have an effect if they are larger than 0.

### LINKED ISSUE #486: CS-BOLD-017 Informational 7.10: Misleading Function Names in Zapper
In GasCompZapper and LeverageLSTZapper, the same function names are used as in their counterparts WETHZapper and LeverageWETHZapper.
For example, `adjustTroveWithRawETH()`: In WETHZapper the function name is accurate, as raw ETH can be wrapped or unwrapped to be used in the function. However, in GasCompZapper the function never operates with raw ETH.

bingen:
Closes #484
Closes #486
Closes #479


contracts/src/Zappers/GasCompZapper.sol
```diff
@@ -191,6 +191,8 @@ contract GasCompZapper is AddRemoveManagers, LeftoversSweep {
         bool _isDebtIncrease,
         InitialBalances memory _initialBalances
     ) internal returns (address) {
+        require(!_isDebtIncrease || _boldChange > 0, "GCZ: Increase bold amount should not be zero");
+
```
danielattilasimon, line 195: Now it's consistent with WETHZapper, but it still seems inconsistent between coll & debt to me (this goes for WETHZapper, too):

- `isDebtIncrease = true` requires `_boldChange > 0`
- `isCollIncrease = true` doesn't require `_collChange > 0`, in fact you _must_ pass `_isCollIncrease = true` along with `_collChange = 0` if you are add-manager but not remove-manager and want to repay some debt but not add any collateral.

Wouldn't this be better?
```solidity
if ((!_isCollIncrease && _collChange > 0) || (_isDebtIncrease && _boldChange > 0)) {
    receiver = _requireSenderIsOwnerOrRemoveManagerAndGetReceiver(_troveId, owner);
}

if ((_isCollIncrease && _collChange > 0) || (!_isDebtIncrease && _boldChange > 0)) {
    _requireSenderIsOwnerOrAddManager(_troveId, owner);
}
```

Then we could drop the `require(!_isDebtIncrease || _boldChange > 0)`.

bingen: For now I put those conditions in a common `BaseZapper` contract... At least we ensure consistency :sweat_smile:

danielattilasimon: Yeah, that looks good!

contracts/src/Zappers/WETHZapper.sol
```diff
@@ -224,19 +205,36 @@ contract WETHZapper is AddRemoveManagers, LeftoversSweep {
...
         // return leftovers to user
-        _returnLeftovers(WETH, boldToken, _initialBalances);
+        uint256 currentBoldBalance = boldToken.balanceOf(address(this));
+        if (currentBoldBalance > _initialBalances.boldBalance) {
+            boldToken.transfer(_initialBalances.sender, currentBoldBalance - _initialBalances.boldBalance);
+        }
+
+        uint256 currentCollBalance = WETH.balanceOf(address(this));
+        if (currentCollBalance > ethToSend + _initialBalances.collBalance) {
+            ethToSend = ethToSend + currentCollBalance - ethToSend - _initialBalances.collBalance;
```
danielattilasimon, line 227: In other words: `currentCollBalance - _initialBalances.collBalance`.
Why the `ethToSend - ethToSend`?

danielattilasimon: Moreover: in the case of BOLD, we treat newly minted BOLD differently from "leftover" BOLD, in that leftovers are sent back to the original `msg.sender` while newly minted BOLD is sent to the receiver. This makes sense.

However, in the case of (W)ETH, all of it is sent to the receiver, even leftovers. Can there be leftovers in the first place? I don't think so, which would make this a non-issue, but in that case the code is more complicated than it needs to be.

bingen: :man_facepalming:

bingen: Should be fixed now.

danielattilasimon (approved):
Looks good now! 👍

PR #521, opened 2024-10-17
feat: Add Hybrid (Curve + UniV3) exchange for Zappers

### LINKED ISSUE #517: Allow multiple hops swaps for Leverage
It’s not likely that we’ll have plenty of Liquidity for WETH-BOLD and all LST-BOLD pairs, so we need to be able to do things like BOLD-USDC-Collateral.

bingen:
Closes #517


contracts/src/Zappers/Modules/Exchanges/HybridCurveUniV3Exchange.sol
```diff
@@ -0,0 +1,203 @@
...
+        _returnLeftovers(initialBalances);
+
+        return boldAmount;
+    }
+
+    function _setHybridExchangeInitialBalances(InitialBalances memory initialBalances) internal view {
+        initialBalances.tokens[0] = boldToken;
+        initialBalances.tokens[1] = USDC;
+        initialBalances.tokens[2] = WETH;
+        initialBalances.tokens[3] = collToken;
```
danielattilasimon, line 200: I guess it doesn't hurt, but seems redundant to add WETH twice, if it's the collToken of the branch.

contracts/src/Zappers/Modules/Exchanges/HybridCurveUniV3Exchange.sol
```diff
@@ -0,0 +1,163 @@
...
+    function swapFromBold(uint256 _boldAmount, uint256 _minCollAmount) external returns (uint256) {
+        InitialBalances memory initialBalances;
+        _setHybridExchangeInitialBalances(initialBalances);
+
+        // Curve
+        ICurveStableswapNGPool curvePoolCached = curvePool;
+        boldToken.transferFrom(msg.sender, address(this), _boldAmount);
+        boldToken.approve(address(curvePoolCached), _boldAmount);
+
+        uint256 curveUsdcAmount = curvePoolCached.exchange(int128(BOLD_TOKEN_INDEX), int128(USDC_INDEX), _boldAmount, 0);
```
RickGriff, line 78: I guess if this gets sandwiched the Curve pool may return much less USDC than desired, but ultimately the `_minCollAmount` ensures the caller gets what they want (or else it reverts)

contracts/src/Zappers/Modules/Exchanges/HybridCurveUniV3Exchange.sol
```diff
@@ -0,0 +1,163 @@
...
+
+        collToken.safeTransferFrom(msg.sender, address(this), _collAmount);
+        collToken.approve(address(uniV3RouterCached), _collAmount);
+
+        // See: https://docs.uniswap.org/contracts/v3/guides/swaps/multihop-swaps
+        bytes memory path;
+        if (address(WETH) == address(collToken)) {
+            path = abi.encodePacked(WETH, feeUsdcWeth, USDC);
+        } else {
+            path = abi.encodePacked(collToken, feeWethColl, WETH, feeUsdcWeth, USDC);
```
RickGriff, line 126: Is this expected to be the best path (i.e. lowest slippage) specifically for RETH and WSTETH? Though maybe we're not trying to overly optimize that.

RickGriff (approved):
Looks good, as far as I can see.

contracts/src/Zappers/Modules/Exchanges/HybridCurveUniV3Exchange.sol
```diff
@@ -0,0 +1,163 @@
...
+
+        // Curve
+        ICurveStableswapNGPool curvePoolCached = curvePool;
+        boldToken.transferFrom(msg.sender, address(this), _boldAmount);
+        boldToken.approve(address(curvePoolCached), _boldAmount);
+
+        uint256 curveUsdcAmount = curvePoolCached.exchange(int128(BOLD_TOKEN_INDEX), int128(USDC_INDEX), _boldAmount, 0);
+
+        // Uniswap
+        ISwapRouter uniV3RouterCached = uniV3Router;
```
danielattilasimon, line 81: Is there a point in caching immutables? Why cache the router specifically, but not `USDC` or `WETH`?

bingen: Nope, that’s an overlook. Thanks!

contracts/src/Zappers/Modules/Exchanges/HybridCurveUniV3Exchange.sol
bingen: Yes, good point. I thought it would add unnecessarily complexity to try to fail earlier if we detect that the amount obtained in the first step is not going to be enough.

contracts/src/Zappers/Modules/Exchanges/HybridCurveUniV3Exchange.sol
bingen: Yes, that’s the assumption for now. Eventually we’ll create the dynamic version suggested by @danielattilasimon where the route can be passed as an argument.

PR #523, opened 2024-10-18
Upfrontfee on joining batch

bingen:


contracts/src/BorrowerOperations.sol
```diff
@@ -993,14 +993,14 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
         newBatchTroveChange.newWeightedRecordedDebt =
             (vars.newBatch.entireDebtWithoutRedistribution + vars.trove.entireDebt) * vars.newBatch.annualInterestRate;
 
+        // Apply upfront fee
         // We may check the old rate to see if it’s different than the new one, but then we should check the
         // last interest adjustment times to avoid gaming. So we decided to keep it simple and account it always
         // as a change. It’s probably not so common to join a batch with the exact same interest rate.
```
RickGriff, line 999: This makes sense, but I wonder if it's a bit complex for someone reading the code fresh. For the whole comment maybe we can just write:

"An upfront fee is always charged upon joining a batch to ensure that borrowers can not game the fee logic and gain free interest rate updates (e.g. if they also manage the batch they joined)"

RickGriff (approved):
Looks good!

danielattilasimon (approved):
Looks good! I fixed some tests.

PR #526, opened 2024-10-20
feat: events when setting add/remove managers

### LINKED ISSUE #487: CS-BOLD-029 Informational 7.11: Missing Events
The functions `AddRemoveManager._setAddManager()` and `AddRemoveManager._setRemoveManager()` do not emit events when the manager is set. This can make it difficult to track changes to the manager.
Since Trove NFTs can be sent to other addresses, the missing events make it hard for the owner to track the current add and remove manager addresses and could lead to users receiving Trove NFTs without knowing the current manager addresses

danielattilasimon:
Fixes #487


bingen (approved):
Looks good!

PR #527, opened 2024-10-20
fix: Allow borrowing if it brings TCR back up to CCR

### LINKED ISSUE #511: Recon M-04: Borrowing from branches can be disabled by one whale or early depositor
Below CCR, we disallow some borrower ops. A whale could bring TCR ~=CCR and DoS them.

bingen:
Fixes #511

### DIFF
--- a/contracts/src/BorrowerOperations.sol
+++ b/contracts/src/BorrowerOperations.sol
@@ -132,7 +132,6 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
     error TroveNotZombie();
     error UpfrontFeeTooHigh();
     error BelowCriticalThreshold();
-    error BorrowingNotPermittedBelowCT();
     error ICRBelowMCR();
     error RepaymentNotMatchingCollWithdrawal();
     error TCRBelowCCR();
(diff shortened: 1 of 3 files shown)


danielattilasimon:
So all along, opening a Trove below CCR was okay as long as it would tip the TCR above CCR? But borrowing against an existing Trove was a no-go. Well, it's good that we're making it consistent then.

bingen:
No, I think it was already consistent, see here:
https://github.com/liquity/bold/pull/527/files#diff-d30cc87ed051c6909d6e9b48a0d719549c2272b1418110c6f55618d5ec6d9448L322

The aim is to fix the issue brought up by Alex, where a whale may block borrowing in general (both new and existing troves) by being able to place the system slightly below CCR.

danielattilasimon:
Ah, I see, I missed the check being removed from `openTrove()`. Then we'll also need to change the `openTrove()` handler in the invariant test.

contracts/src/BorrowerOperations.sol
```diff
@@ -1350,26 +1347,14 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
         }
     }
 
-    function _requireNotBelowCriticalThreshold(uint256 _price) internal view {
-        if (_checkBelowCriticalThreshold(_price, CCR)) {
-            revert BelowCriticalThreshold();
```
danielattilasimon, line 1355: This makes `error BelowCriticalThreshold()` unused, I believe.

bingen: Oh, right, I removed `BorrowingNotPermittedBelowCT` but forgot this one. Thanks for fixing it!

bingen:
And thanks for fixing the invariant tests too!

danielattilasimon (approved):
Looks good now!

PR #528, opened 2024-10-20
CS-BOLD-023: Trove validation in urgent redemptions

### LINKED ISSUE #471: CS-BOLD-023 Low 5.10: Missing Validation of Troves in Urgent Redemptions
The function `TroveManager.urgentRedemption` does not validate if a trove ID exists or if the trovehas any debt to redeem. This can lead to unnecessary gas consumption when a trove has no debt or has been previously redeemed.
Further it accepts Troves that have been closed or never existed, causing writes to the following storage locations by `_applySingleRedemption`:
• `rewardSnapshots[_troveId].coll = L_coll`
• `rewardSnapshots[_troveId].boldDebt = L_boldDebt`
• `Troves[_troveId].lastDebtUpdateTime = uint64(block.timestamp)`

The dirty memory locations cannot cause invalid state modifications within the protocol, but they are returned by getter functions and can be read by external applications.

RickGriff:
Fixes #471

### DIFF
--- a/contracts/src/TroveManager.sol
+++ b/contracts/src/TroveManager.sol
@@ -455,7 +455,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         _sendGasCompensation(activePoolCached, msg.sender, totals.ETHGasCompensation, totals.collGasCompensation);
     }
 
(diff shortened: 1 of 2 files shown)


danielattilasimon (approved):
Looks good to me too!

PR #529, opened 2024-10-21
chore: Fix code comments

### LINKED ISSUE #469: CS-BOLD-007 Low 5.8: Incorrect Code Comments
1. The code comments of the function `BorrowerOperations._mintBatchManagementFeeAndAccountForChange()` state that the arithmetic is done in two steps to avoid overflow. However, the arithmetic could only underflow from
the decrease.
2. The natspec in the function `BorrowerOperations._openTrove()` mentions a so-called composite debt that includes bold gas compensation. The concept of composite debt was used in Liquity V1, but is not present in Liquity V2, where the gas compensation is charged in WETH instead.
3. Similar to 2., code comments in `BorrowerOperations._applyUpfrontFee()` mention the same deprecated concept of composite debt.
4. The code comments in `TroveManager.redeemCollateral()` write the word "proportinally" instead of "proportionally".
5. The code comments in `TroveManager.redeemCollateral()` state that troves are redeemed based on their collateral ratio. However, redemption order is now actually based on the interest rate.
6. The natspec above the function `LiquityMath._decPow()` mentions two functions `TroveManager._calcDecayedBaseRate` and `CommunityIssuance._getCumulativeIssuanceFraction` that are not part of the repository.
7. The natspec above the public variable in the `StabilityPool.sortedTroves` mentions that the state variable is used for liquidations, but it is not used anywhere.
8. The code comments above the struct `TroveManager.Batch` state that the collateral is shared between troves in a batch, but collateral is kept separately.
9. The code comments above the state variable `MainnetPriceFeedBase.priceFeedDisabled` state that it should be removed after shutdown logic is implemented, but the code relies on it to return the fallback price.

#. The comments in `TroveManager._urgentRedeemCollateralFromTrove()` state that collateral has to be capped as the CR can be below 100% for urgent redemptions. However, due to the bonus multiplier, CR < 101% already requires capping the reward.
Additionally, there is a typo in the code:
• In `BorrowerOperations.adjustUnredeemableTrove`, `batchAnnualInterestRate` is written as
`batchAnnualInteresRate`.

bingen:
Fixes #469

### DIFF
--- a/contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
+++ b/contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
@@ -10,7 +10,7 @@ import "../BorrowerOperations.sol";
 // import "forge-std/console2.sol";
 
 abstract contract MainnetPriceFeedBase is IPriceFeed, Ownable {
-    // Dummy flag raised when the collateral branch gets shut down.
+    // Flag raised when the collateral branch gets shut down.
     bool priceFeedDisabled;
 
     // Last good price tracker for the derived USD price
--- a/contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
+++ b/contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
@@ -11,7 +11,6 @@ import "../BorrowerOperations.sol";
 
 abstract contract MainnetPriceFeedBase is IPriceFeed, Ownable {
     // Dummy flag raised when the collateral branch gets shut down.
-    // Should be removed after actual shutdown logic is implemented.
     bool priceFeedDisabled;
 
     // Last good price tracker for the derived USD price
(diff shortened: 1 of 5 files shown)


contracts/src/PriceFeeds/MainnetPriceFeedBase.sol
```diff
@@ -11,7 +11,6 @@ import "../BorrowerOperations.sol";
 
 abstract contract MainnetPriceFeedBase is IPriceFeed, Ownable {
     // Dummy flag raised when the collateral branch gets shut down.
-    // Should be removed after actual shutdown logic is implemented.
```
danielattilasimon, line 14: If the flag is here to stay (as it appears to be) then I would no longer call it a dummy.

contracts/src/TroveManager.sol
```diff
@@ -25,7 +25,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
     address internal gasPoolAddress;
     ICollSurplusPool internal collSurplusPool;
     IBoldToken internal boldToken;
-    // A doubly linked list of Troves, sorted by their sorted by their collateral ratios
+    // A doubly linked list of Troves, sorted by their sorted by their interest rate
```
danielattilasimon, line 28: While we're at it, we might as well fix the double "sorted by their".

danielattilasimon (approved):
Looks good!

PR #530, opened 2024-10-21
chore: TODOs, forge fmt...

### LINKED ISSUE #480: CS-BOLD-025 Informational 7.4:
We have identified the following remaining code comments that should be removed before deployment:
1. Several functions in the code have open TODOs.
2. Code for batch redistributions in function `TroveManager._getLatestTroveDataFromBatch()` is commented out.
3. The TroveManager contract has imports for testing imports // import "forge-std/console2.sol";

bingen:
Fixes #480

### DIFF
--- a/contracts/src/ActivePool.sol
+++ b/contracts/src/ActivePool.sol
@@ -164,7 +164,7 @@ contract ActivePool is IActivePool {
     function sendColl(address _account, uint256 _amount) external override {
         _requireCallerIsBOorTroveMorSP();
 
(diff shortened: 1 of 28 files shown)


danielattilasimon (approved):
Looks good, but I would run `forge fmt` again, as there are now a lot of empty lines where the `console2.sol` import used to be, and that's removed by the formatter.

bingen:
> but I would run `forge fmt` again

Absolutely, I was planning to do that, but I forgot :sweat_smile: 
Thanks!

PR #531, opened 2024-10-21
Upgrade to Solidity 0.8.24 and remove some floating pragma

### LINKED ISSUE #482: CS-BOLD-026 Informational 7.6: Floating Pragma
Liquity uses a floating pragma in some of the contract files (i.e. AddressesRegistry.sol). It is considered best practice to lock the Solidity version, to ensure the contracts are tested with the same compiler version as they are deployed.

bingen:
Fixes #482


danielattilasimon (approved):
Looks good!

PR #534, opened 2024-10-22
fix: Gas optimizations

### LINKED ISSUE #483: CS-BOLD-027 Informational 7.7: Gas Optimizations
1. The function `ActivePool.mintBatchManagementFeeAndAccountForChange()` is protected by the `_requireCallerIsBOorTroveM()` modifier, but is only called by the TroveManager contract and never the BorrowOperations.
2. The interest calculation in `ActivePool.calcPendingAggInterest()` could short-escape and return 0 if the last update has been at the current timestamp. Similarly, the function `ActivePool.calcPendingAggBatchManagementFee()` could short-escape early.
3. The function `TroveManager._getLatestTroveDataFromBatch()` reads the entire structure batch from storage, but only uses one value from it.
4. The modifier `CollateralRegistry._requireBoldBalanceCoversRedemption()` asserts that the user's balance is smaller than the total supply, but this invariant is already enforced by OZ's token implementation.
5. The  function `TroveManager._redistributeDebtAndColl` can  skip  the  call `sendCollToDefaultPool` when the collateral to redistribute is zero.
6. The function `TroveManager._removeTroveSharesFromBatch()` subtracts the upfront fee from the new trove debt. However, the upfront fee is always zero when positions are closed or liquidated.
7. The modifier `UniV3Exchange._requireCallerIsUniV3Router()` is never used.
8. The function `UniV3Exchange.uniswapV3SwapCallback()` is never used.
9. The `BorrowerOperations.adjustTroveInterestRate` function only charges an upfront fee when the new interest rate is not equal to the previous rate, but this is not possible due to the check in `_requireAnnualInterestRateIsNew`.
10. The immutable `StabilityPool.sortedTroves` is never used.
11. The modifier `BorrowerOperations._requireIsShutDown()` is never used.
12. The function `WSTETHPriceFeed._fetchPrice()` calls `wstETH.stEthPerToken()`, which in turn calls `stETH.getPooledEthByShares`. It would be cheaper to call stETH directly.
13. The TroveManager contract could define state variables only set once in the constructor as immutable: collateralRegistry, sortedTroves, boldToken, collSurplusPool, gasPoolAddress, stabilityPool, borrowerOperations, troveNFT.
14. The BorrowerOperations contract could define state variables only set once in the constructor as immutable: troveManager, gasPoolAddress, collSurplusPool, sortedTroves, boldToken.
15. The ActivePool contract could define state variables only set once in the constructor as immutable: interestRouter, boldToken.

bingen:
Mostly, use of immutable variables and removal of unsued functions / variables.

Fixes #483

### DIFF
--- a/contracts/src/BorrowerOperations.sol
+++ b/contracts/src/BorrowerOperations.sol
@@ -1249,12 +1249,6 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
         }
     }
 
-    function _requireIsShutDown() internal view {
-        if (!hasBeenShutDown) {
-            revert NotShutDown();
-        }
-    }
-
     function _requireNonZeroAdjustment(TroveChange memory _troveChange) internal pure {
         if (
             _troveChange.collIncrease == 0 && _troveChange.collDecrease == 0 && _troveChange.debtIncrease == 0
--- a/contracts/src/BorrowerOperations.sol
+++ b/contracts/src/BorrowerOperations.sol
@@ -116,7 +116,6 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
     }
 
     error IsShutDown();
-    error NotShutDown();
     error TCRNotBelowSCR();
     error ZeroAdjustment();
     error NotOwnerNorInterestManager();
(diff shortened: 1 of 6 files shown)


contracts/src/BorrowerOperations.sol
```diff
@@ -1249,12 +1249,6 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
         }
     }
 
-    function _requireIsShutDown() internal view {
-        if (!hasBeenShutDown) {
-            revert NotShutDown();
```
danielattilasimon, line 1254: I guess we can remove `error NotShutDown()` from `BorrowerOperations`.

danielattilasimon (approved):
Looks good!

PR #535, opened 2024-10-22
chore: Make comment about skipped redemptions more precise

bingen:


danielattilasimon (approved):
Looks good!

It made me think: could we have made the threshold dynamic (i.e. slightly higher than 100%) based on the redemption rate, now that the redemption fee is left in the Trove? Anyway, I wouldn't change it now, just a thought.

bingen:
Yes, I thought so too, but I agree I wouldn’t touch it now.

PR #538, opened 2024-10-23
fix: Check for remaining bold redemption amount early

### LINKED ISSUE #537: Debt applied after branch shut down
Once a branch is shut down, `applyPendingDebt()` is no longer allowed. I guess this is because interest & fees stop accruing once a branch is shut down, so compounding makes no sense. (But then how's a batch manager supposed to mint the batch management fees that they're owed)?
However, it seems this can be circumvented by urgently redeeming 0 BOLD from a Trove, which is basically the same as `applyPendingDebt()`: it mints pending batch management fees and updates the Trove's recorded debt, but does nothing else. This is because the remaining BOLD amount is checked after one iteration of redemption already takes place:
```
function urgentRedemption(uint256 _boldAmount, uint256[] calldata _troveIds, uint256 _minCollateral) external {
    // ... init stuff ...

    uint256 remainingBold = _boldAmount;
    for (uint256 i = 0; i < _troveIds.length; i++) {
        // ... redemption effects ...

        remainingBold -= singleRedemption.boldLot;
        if (remainingBold == 0) break;                 // !!! why check this late?
    }

    // ... global effects ...
}
```
For reference, normal redemption wouldn't touch any Troves in the same situation.

bingen:
To avoid applying debt to a trove after shutdown by passing zero as `_boldAmount`.

Fixes #537

### DIFF
--- a/contracts/src/TroveManager.sol
+++ b/contracts/src/TroveManager.sol
@@ -849,6 +849,8 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
 
         uint256 remainingBold = _boldAmount;
         for (uint256 i = 0; i < _troveIds.length; i++) {
+            if (remainingBold == 0) break;
+
             SingleRedemptionValues memory singleRedemption;
             singleRedemption.troveId = _troveIds[i];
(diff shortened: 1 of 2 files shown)


danielattilasimon (approved):
Looks good!

PR #540, opened 2024-10-23
fix: Apply SP error fix to scale updates

### LINKED ISSUE #518: SP: Should we apply the underflow fix to the case where we update the scale?
Address the TODO from #438

bingen:
Fixes #518


danielattilasimon (approved):
Looks good to me, and it has been running on the fuzzer for a week without issues, but I would wait for @RickGriff's approval, too.

bingen:
Replaced by #552

PR #546, opened 2024-10-30
fix: Fix wrong uni V3 fee in Zappers hybrid exchange

bingen:


danielattilasimon (approved):
Oops, good catch!

bingen:
It was Torgin from ChainSecurity :wink:

PR #547, opened 2024-10-30
fix: Send leftovers to receiver in Zappers lever up and down

bingen:
Also, rename `_setInitialBalances` in Zappers


danielattilasimon (approved):
Good point! Tokens should not be leaking towards the manager. If they must leak, it should be towards the receiver.

PR #560, opened 2024-11-01
fix: Make sure a fully redeemed trove ends up with zero shares in a batch

### LINKED ISSUE #416: Batched Troves can end up with (tiny) non-zero shares after a full redemption
Example of this happening on the testnet:
https://testnet.liquity.org/tx/0x05eb9fc0c27fcb5c43bce0e2ac91df49c186a5f327999fc4dae8127073d72b98/logs

(Look for `BatchedTroveUpdated` events).

bingen:
Closes #416

### DIFF
--- a/contracts/src/TroveManager.sol
+++ b/contracts/src/TroveManager.sol
@@ -1264,6 +1264,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         // Push the trove's id to the Trove list
         TroveIds.push(_troveId);
 
+        assert(_troveChange.debtIncrease > 0); // TODO: remove before deployment
         _updateBatchShares(_troveId, _batchAddress, _troveChange, _troveChange.debtIncrease, _batchColl, _batchDebt, true);
 
         uint256 newTotalStakes = totalStakes + newStake;
@@ -1517,6 +1518,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         uint256 newStake = _updateStakeAndTotalStakes(_troveId, _newTroveColl);
 
         // Batch
+        assert(_newTroveDebt > 0); // TODO: remove before deployment
         _updateBatchShares(_troveId, _batchAddress, _troveChange, _newTroveDebt, _newBatchColl, _newBatchDebt, true);
 
         _movePendingTroveRewardsToActivePool(
@@ -1572,6 +1574,7 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         Troves[_troveId].coll = _newTroveColl;
 
         if (_batchAddress != address(0)) {
+            assert(_newTroveDebt > 0); // TODO: remove before deployment
             _updateBatchShares(_troveId, _batchAddress, _troveChange, _newTroveDebt, _newBatchColl, _newBatchDebt, true);
 
             emit BatchUpdated({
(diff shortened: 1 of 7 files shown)


contracts/src/test/TestContracts/InvariantsTestHandler.t.sol
```diff
@@ -1535,6 +1535,7 @@ contract InvariantsTestHandler is Assertions, BaseHandler, BaseMultiCollateralTe
         v.batchManager = _batchManagerOf[i][v.troveId];
         v.batchManagementFee = v.c.troveManager.getLatestBatchData(v.batchManager).accruedManagementFee;
         v.t = v.c.troveManager.getLatestTroveData(v.troveId);
+        vm.assume(v.t.entireDebt > 0); // skip troves if there’s no pending debt
```
danielattilasimon, line 1538: This goes against the spirit of the invariant tests. If we know an operation is expected to fail, we go ahead anyway, and make sure that it does fail, and for the right reason.

danielattilasimon: Also, when an operation does succeed, we should ensure that all preconditions are met. We are adding a new precondition (debt > 0) which means a corresponding check should be added.

contracts/src/test/TestContracts/InvariantsTestHandler.t.sol
```diff
@@ -1546,6 +1545,7 @@ contract InvariantsTestHandler is Assertions, BaseHandler, BaseMultiCollateralTe
         try v.c.borrowerOperations.applyPendingDebt(v.troveId, v.lowerHint, v.upperHint) {
             // Preconditions
             assertTrue(v.wasOpen, "Should have failed as Trove wasn't open");
+            assertGtDecimal(v.t.entireDebt, 0, 18, "Should have failed as debt was zero");
```
bingen, line 1548: Ah, nice, thanks!

contracts/src/TroveManager.sol
```diff
@@ -1263,7 +1264,8 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         // Push the trove's id to the Trove list
         TroveIds.push(_troveId);
 
-        _updateBatchShares(_troveId, _batchAddress, _troveChange, _batchColl, _batchDebt, true);
+        assert(_troveChange.debtIncrease > 0); // TODO: remove before deployment
```
RickGriff, line 1267: Actually, I wondered if we could have fixed the original issue with a check that batch debt always increases by the same amount as the caller's BOLD balance (or, at least within some low tolerance). e.g. enforcing the core borrowing invariant.

Though I think we'd still need this PR for redistributions.

contracts/src/TroveManager.sol
```diff
@@ -1794,11 +1801,20 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
                 batches[_batchAddress].totalDebtShares = currentBatchDebtShares + batchDebtSharesDelta;
             } else if (debtDecrease > 0) {
                 // Subtract debt
-                batchDebtSharesDelta = currentBatchDebtShares * debtDecrease / _batchDebt;
+                // We make sure that if final trove debt is zero, shares are too (avoinding rounding issues)
```
RickGriff, line 1804: typo here: avoinding -> avoiding

contracts/src/TroveManager.sol
```diff
@@ -1794,11 +1801,20 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
...
+                // This can only happen from redemptions, as otherwise we would be using _removeTroveSharesFromBatch
+                // In redemptions we don’t do that because we don’t want to kick the trove out of the batch
```
RickGriff, line 1806: I guess we actually could (though it would be strange UX).

RickGriff (approved):
Looks good!

danielattilasimon (approved):
Looks good to me to, and fixes some rare counterexamples when urgently redeeming Troves having microscoping debts multiple times within the same call.

Issue #488, opened 2024-10-09
CS-BOLD-030 Informational 7.12: Out-of-gas May Lead to Shutdown


bingen:
The MainnetPriceFeedBase shuts down the collateral branch when the static call to latestRoundData() reverts. The fallback uses a try-catch block as follows:

```
// Secondly, try to get latest price data:
try _aggregator.latestRoundData() returns (uint80 roundId, int256 answer, uint256, /* startedAt */ uint256 updatedAt, uint80 /* answeredInRound */) {
    // If call to Chainlink succeeds, return the response and success = true
    chainlinkResponse.roundId = roundId;
    chainlinkResponse.answer = answer;
    chainlinkResponse.timestamp = updatedAt;
    chainlinkResponse.success = true;
    return chainlinkResponse;
} catch {
    // If call to Chainlink aggregator reverts, return a zero response with success = false
    return chainlinkResponse;
}
```

There are two cases in which the external call to Chainlink reverts: Either Chainlink explicitly reverts, or the call runs out of gas. As such, the catch statement (which triggers branch shutdown) can be executed if the call to Chainlink runs out of gas. However, this will only have an effect if there is enough gas left to execute the shutdown logic. The call to Chainlink will receive 63/64 of all available gas. After reverting due to out of gas, 1/64 will be left. This means that if the shutdown logic consumes 64 times less gas than the Chainlink's latestRoundData(), the branch can unintentionally be shutdown even though the Chainlink oracle has not failed. The code has no access-control, so anyone can call fetchPrice() with any amount of gas.

At the time of writing, a call to `latestRoundData` appears to use approximately 11000 gas units. `1/64 * 11000 = 172`, which is by far not enough to execute the branch shutdown. Given these conditions, the attack is currently not feasible.

Note that an attacker can prewarm storage slots and addresses to reduce the cost of the remainder of the execution. Further, note that the gas consumption of Chainlink might increase in the future (as the contracts are upgradeable). Additionally, the gas cost of Ethereum opcodes could change in the future.

bingen:
Fixed in #569

PR #569, opened 2024-11-05
Eth price fallback 2 fixes

RickGriff:
This PR:

- Implements CS's recommendation for using the `max(STETH-USD,ETH-USD)` for redemptions only if those prices differ by 1%, otherwise take the min:

image

- Stops out-of-gas errors in external calls to Chainlink and LST causing a shutdown. Fixed by ensuring the remaining post-call gas is > 1/64 the pre-call gas, which guarantees an OOG didn't occur ( CS-030:  https://github.com/liquity/bold/issues/488)
- Makes CompositePriceFeed an abstract contract (Coinspect BOLD-06)


bingen (approved):
Looks good!

PR #571, opened 2024-11-05
feat: Add Governance deployment script

bingen:
Also, use “hybrid” zapper


contracts/src/scripts/DeployGovernance.s.sol
```diff
@@ -0,0 +1,204 @@
...
+        curvePool = ICurveStableswapNG(
+            curveFactory.deploy_plain_pool(
+                "BOLD-USDC", "BOLDUSDC", _coins, 200, 1000000, 50000000000, 866, 0, _asset_types, _method_ids, _oracles
+            )
+        );
+
+        gauge = ILiquidityGauge(curveFactory.deploy_gauge(address(curvePool)));
+
+        curveV2GaugeRewards =
+            new CurveV2GaugeRewards(address(_governance), address(boldToken), address(lqty), address(gauge), DURATION);
```
danielattilasimon, line 200: For `CurveV2GaugeRewards` to work, BOLD has to be added as a reward token to the gauge. In ChickenBonds, we made this work by deploying the gauge through the "Gauge Manager Proxy" (https://etherscan.io/address/0xd05Ad7fb0CDb39AaAA1407564dad0EC78d30d564#code), which allows the deployer to call `add_reward()` indirectly, through the proxy itself. I don't know if there's a Sepolia equivalent.

bingen: This is supposed to be easier now:
https://docs.curve.fi/curve_dao/ownership-proxy/GaugeProxy/?h=gauge#permissionless-rewards

But I haven’t managed to make it work on Sepolia yet.

PR #578, opened 2024-11-07
fix: Simplify close from coll params in Zappers

bingen:


bpierre (approved):
Nice :sparkles:

danielattilasimon (approved):
Looks good! So there was a `receiver` param but it was unused?

bingen:
It was used as a var, not as a param. So damn ugly. And there’s no even the excuse of the stack too deep issue.

PR #596, opened 2024-11-18
feat: Add Zappers Hybrid exchange helpers

bingen:


contracts/src/scripts/DeployLiquity2.s.sol
```diff
@@ -893,6 +907,7 @@ contract DeployLiquity2Script is DeployGovernance, UniPriceConverter, StdCheats,
                 string.concat('"boldToken":"', address(deployed.boldToken).toHexString(), '",'),
                 string.concat('"hintHelpers":"', address(deployed.hintHelpers).toHexString(), '",'),
                 string.concat('"multiTroveGetter":"', address(deployed.multiTroveGetter).toHexString(), '",'),
+                string.concat('"exchangeHelpers":"', address(deployed.exchangeHelpers).toHexString(), '",'),
```
danielattilasimon, line 910: Should we add the address of the newly deployed `exchangeHelpers` to the existing manifest at [contracts/addresses/11155111.json](https://github.com/liquity/bold/blob/main/contracts/addresses/11155111.json)?

contracts/src/Zappers/Interfaces/IExchangeHelpers.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import "openzeppelin-contracts/contracts/token/ERC20/IERC20.sol";
+
+interface IExchangeHelpers {
+    function getCollFromBold(uint256 _boldAmount, IERC20 _collToken, uint256 _desiredCollAmount) external /* view */ returns (uint256, uint256);
```
danielattilasimon, line 8: We should name the return values, since it's not obvious from the name of the function what the elements of the tuple hold.

contracts/src/Zappers/Modules/Exchanges/HybridCurveUniV3ExchangeHelpers.sol
```diff
@@ -0,0 +1,93 @@
...
+        bytes memory path;
+        if (address(WETH) == address(_collToken)) {
+            path = abi.encodePacked(USDC, feeUsdcWeth, WETH);
+        } else {
+            path = abi.encodePacked(USDC, feeUsdcWeth, WETH, feeWethColl, _collToken);
+        }
+
+        (uint256 collAmount,,,) = uniV3Quoter.quoteExactInput(path, curveUsdcAmount);
+
+        uint256 slippage = collAmount > _desiredCollAmount ? 0: DECIMAL_PRECISION - collAmount * DECIMAL_PRECISION / _desiredCollAmount;
```
danielattilasimon, line 89: I question the usefulness of `slippage` being calculated in the helper. If this is about calculating the price impact of the trade, the frontend would first have to call `getCollFromBold()` passing some small amount of BOLD as input (e.g. $1 BOLD) to get the current BOLD:COLL "spot price", from which `_desiredCollAmount` can be calculated. Then it'd have to call `getCollFromBold()` a second time, now passing a meaningful `_desiredCollAmount`. However, it's more efficient to call `getCollFromBold()` twice in parallel, as it can be performed in a single multicall, then performing the very simple calculation on the client side.

Is there some other use of `_desiredCollAmount` and `slippage` that I'm missing?

danielattilasimon: Having read the tests, I guess the idea is to use the oracle price of the collateral as a basis for calculating `_desiredCollAmount`, but I have 2 issues with that:
- The oracle price is in USD, so we're ignoring the BOLD:USD price.
- Calling it slippage is IMO misleading, as that normally refers to the difference between the exchange rate as predicted and the exchange rate at which the trade is actually executed. What we're calculating here is quite far removed from that.

danielattilasimon: Okay, now I think it's definitely useful to calculate this when you're leveraging, as any deviation from the oracle price on the downside will eat into your leverage, I'm just not sure:
- if slippage is the right term
- if it needs to be included in the helper (the frontend already has to perform half of the calculation to make it work, by fetching the oracle price and calculating `_desiredCollAmount` accordingly)

bingen: > I guess the idea is to use the oracle price of the collateral as a basis for calculating `_desiredCollAmount`

That’s the initial approach, but this function is agnostic to that. It may be even be a value provided by the user. So, the function how the right price is determined.

I agree that the term slippage is not the best one, I’ll change it. It can also be removed, yes, but I just thought it was more convenient, even for manual calls. Do you think it hurts to have it there?

danielattilasimon: As long as it's optional (i.e. we can safely pass 0) I don't think it hurts. I just wouldn't use it if I was writing a frontend, as it introduces a point of serialization (you have to wait for the result of whatever query you want to use to determine the right price before you can make the call) which would make the frontend less responsive.

PR #599, opened 2024-11-21
fix: add `indexed` to `CollSent` event param `_to`

### LINKED ISSUE #598: CS-BOLD-035 Informational 7.6: Indexed Parameters of Events
The event `CollSent` used by the `CollSurplusPool` and `IStabilityPoolEvents` could denote the parameter `_to` with the keyword `indexed`, to make it easier for external applications to query pending collateral withdrawals.

danielattilasimon:
Fixes #598.


ColinPlatt:
LGTM, LFG?

bingen:
Yes, if the graph/frontend can work without it (by using the token one), I’d rather remove it. But no big deal if we don’t do it.

PR #610, opened 2024-11-25
fix: Receiver usage should be constrained to remove manager

### LINKED ISSUE #600: CS-BOLD-034 Informational 7.4: Delegation Specification
The [documentation](https://web.archive.org/web/20240905124114/https://github.com/liquity/BOLD?tab=readme-ov-file#remove-managers) states that a `receiver` address can be chosen, who receives the collateral drawn by the Remove Manager. It does not mention that the `receiver` also receives the tokens when the owner makes a withdrawal.

The `receiver` receives the tokens no matter who initiates a withdrawal, whenever it is set. The docs could be updated to explicitly reflect this.

bingen:
Fixes #600


danielattilasimon (approved):
Looks good!

PR #633, opened 2024-12-03
fix: empty function being called in WETHPriceFeed

bingen:


danielattilasimon (approved):
Oh! So this was rendering the check `assert(priceSource == PriceSource.primary)` in the constructor useless?

PR #634, opened 2024-12-04
fix: Make sure safeTransfer is used for collateral tokens

bingen:
For consistency, besides, although planned collateral tokens seem safe, it makes the code more reusable (thinking of forks, for instance)


danielattilasimon (approved):
Makes sense!

PR #654, opened 2024-12-13
Fix Sepolia deployment

bingen:
- Use LQTY mock with wildcard spender for Sepolia deployments
- Update paths on scripts imports


contracts/addresses/11155111.json
```diff
@@ -1,90 +1,82 @@
...
+      "wethZapper": "0x0000000000000000000000000000000000000000",
+      "gasCompZapper": "0xe6c3e2ae319ee41f97c26d8e604beadee7769bdb",
+      "leverageZapper": "0x07c07801507f7c8f47499556828925013a0bbad4",
+      "collToken": "0xd854bf0fb95aee6a18f2f1aecac7df3a705d7a79"
     }
   ],
   "governance": {
-    "LQTYToken": "0x29f6694517cf47208f126da48a32ecbbc8e41e2d",
-    "LUSD": "0x7da678bdcbb2eb6c22d5871735536dcebc4e0dd7",
-    "LQTYStaking": "0xb2062bb31128c4ce026d4a80e90fdf222e8809e5",
```
bpierre, line 79: I made LQTYStaking [optional](https://github.com/liquity/bold/pull/655/files#diff-0d00d8e31ce0d61333fa6c8f7afb540d45ef611a95a14df31ed4b6d006bd6e75R45) in the manifest-to-app-env script (I get it from the Governance contract and fill it manually now), but I was wondering removing it was intentional?

bingen: Definitely not. I have to check what happened here. I saw there’s also 2 constants changed, and I don’t think I touched that.

danielattilasimon: You changed the constants in #562, which was merged last week. It makes sense that the new deployment manifest would get the updated values.

danielattilasimon: As for `LQTYStaking` being removed: my guess is it was only ever included manually. I don't see it ever being present in the code function that generates the governance manifest:
https://github.com/liquity/bold/blob/f8e21aee9bf28b673cd165ccca9a70af9697e764/contracts/scripts/DeployGovernance.s.sol#L235-L249

bingen: > You changed the constants in https://github.com/liquity/bold/pull/562, which was merged last week.
Ah, right, in my memory this was an old change :sweat_smile: 

> As for LQTYStaking being removed: my guess is it was only ever included manually. 
Good point again! Actually, I have a very rudimentary script that pulls addresses from those auto-generated `run-*.json` files when the script is interrupted for some reason and our final manifest doesn’t get created. There I just pull any created contract, so it becomes more verbose.

Thanks for clarifying it!

bingen:
> Can we move the scripts from scripts => script?

Ah, yes, sure. I just moved it as it was inside `src`, I didn’t realize that `script` was the proper one.
Thanks!

danielattilasimon (approved):
Looks good to me! Actually, I would merge this ASAP because currently a bunch of scripts are broken on main, which this fixes. Locally, I'm already working on top of this branch 😅

PR #657, opened 2024-12-16
fix: More granular events for RETH oracle failure

bingen:


RickGriff (approved):
Looks good! Would just update the top comment to:

" If either the ETH-USD feed or exchange rate is down, shut down and switch to the last good price seen by the system since we need both for primary and fallback price calcs"

(it was out of date before this PR)

PR #659, opened 2024-12-18
fix: unreasonable `Governance` config

danielattilasimon:


bingen (approved):
Looks good, thanks!

contracts/script/DeployGovernance.s.sol
```diff
@@ -218,7 +215,6 @@ contract DeployGovernance is Script, Deployers {
                 string.concat(
                     '"UNREGISTRATION_THRESHOLD_FACTOR":"', uint256(UNREGISTRATION_THRESHOLD_FACTOR).toString(), '",'
                 ),
-                string.concat('"REGISTRATION_WARM_UP_PERIOD":"', uint256(REGISTRATION_WARM_UP_PERIOD).toString(), '",'),
```
danielattilasimon, line 221: The constants are commented out of the manifest anyway, should we put them back?

PR #663, opened 2024-12-20
Chore/multiply

ColinPlatt:


frontend/app/.env
bpierre, line 1: We should revert all the changes in this file (no big deal for the keys leaking, I will generate new ones).

frontend/app/src/comps/Positions/PositionCardLeverage.tsx
```diff
@@ -65,7 +65,7 @@ export function PositionCardLeverage({
               color: "positionContent",
             })}
           >
-            <div>Leverage loan</div>
+            <div>Multiply</div>
```
bpierre, line 68: Do we also call the object “Multiply”? For, we were having the app feature called “Leverage”, and the objects created from it “Leverage loan”.

contracts/lib/V2-gov
bpierre, line 1: We probably want to revert the V2-gov update in this PR.

frontend/app/src/content.tsx
```diff
@@ -275,7 +275,7 @@ export default {
     action: "Next: Summary",
     infoTooltips: {
       leverageLevel: [
-        "Leverage level",
+        "Multiply level",
```
bpierre, line 278: Wondering if we should call the level “Multiplier”?

frontend/app/src/tx-flows/updateLeveragePosition.tsx
```diff
@@ -65,8 +65,8 @@ const stepNames: Record<Step, string> = {
   approveLst: "Approve {tokenName}",
   decreaseDeposit: "Decrease Deposit",
   increaseDeposit: "Increase Deposit",
-  leverDownTrove: "Decrease Leverage",
-  leverUpTrove: "Increase Leverage",
+  leverDownTrove: "Decrease Multiply",
+  leverUpTrove: "Increase Multiply",
```
bpierre, line 69: This sounds a bit like two contradictory verbs to me (decrease & multiply), maybe using multiplier (“Decrease multiplier”) could help?

pnpm-lock.yaml
bpierre, line 1: We should revert this as no dependencies got updated.

bpierre:
:pray:  I left some comments and also added a few missing bits.

PR #664, opened 2024-12-20
Add state needed for calculating voting power to subgraph

danielattilasimon:
We add the necessary state needed for calculating both total and per-user voting power. This could be useful if one wants to know e.g. their share of the total votes. Voting power is internally represented by a pair of `(lqty, offset)` values, from which the voting power at time `t` can be calculated as: `votingPower(t) = lqty * t - offset`.

Also get Governance ABI from forge, just like for other contracts. We can do this now that Governance is a submodule of the repo.

Also fixes an issue where people's LQTY stake could be wrong in the subgraph if someone sent LQTY directly to their UserProxy.

Also, now we have events when the initial initiatives are being registered, so we no longer need the block handler.


subgraph/src/Governance.mapping.ts
```diff
@@ -1,38 +1,24 @@
...
-    initiative.registrant = Address.zero();
-    initiative.save();
-  }
-
+function initializeStats(): GovernanceStats {
   // Create initial stats
   let stats = new GovernanceStats("stats");
   stats.totalLQTYStaked = BigInt.fromI32(0);
+  stats.totalOffset = BigInt.fromI32(0);
   stats.totalInitiatives = 0;
```
danielattilasimon, line 18: Just noticed `totalInitiatives` is never increased. Is the idea to track the number of GovernanceInitiative entities (i.e. only increased, never decreased, not even on unregistration)?

bpierre: No the idea was to track the number of active initiatives, this is something that was missing.

bpierre (approved):
Thanks @danielattilasimon, that’s great!

I deployed this version, and I built on top of it for this PR: https://github.com/liquity/bold/pull/672

Do you think we could add the total voting power somewhere in the graph? Or is it something I could calculate easily with what we have already? So that the app could show a % rather than a unitless number.

danielattilasimon:
> Thanks @danielattilasimon, that’s great!
> 
> I deployed this version, and I built on top of it for this PR: #672
> 
> Do you think we could add the total voting power somewhere in the graph? Or is it something I could calculate easily with what we have already? So that the app could show a % rather than a unitless number.

Both the total voting power and your individual voting power are linear functions of time, but their y-intercepts (called "offset" in the backend and subgraph) are different, hence your % share will change over time even if nobody changes their stake. As such, I would leave it to the frontend to calculate the user's most up-to-date share like this:

```
totalVotingPower(t) = governanceStats.totalLQTYStaked * t - governanceStats.totalOffset
userVotingPower(t) = governanceUser.stakedLQTY * t - governanceUser.stakedOffset
pctShare(t) = userVotingPower(t) / totalVotingPower(t)
```

where `t` is a seconds-based timestamp just like the block timestamp.

bpierre:
@danielattilasimon added in https://github.com/liquity/bold/pull/677, thank you :pray:

PR #672, opened 2024-12-20
App: show live voting power

bpierre:


danielattilasimon (approved):
Looks good! I've left a comment in the other PR how we could show the user's % share of total voting power:
https://github.com/liquity/bold/pull/664#issuecomment-2558324377

PR #681, opened 2025-01-08
Add LICENSE to app

ColinPlatt:


contracts/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "repository": "github:liquity/bold",
   "author": "Liquity AG",
-  "license": "GPL-3.0",
+  "license": "BSL-1.1",
```
danielattilasimon, line 23: BSL is Boost Software License, it should be `BUSL-1.1`:
https://spdx.org/licenses/

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "liquity2",
   "private": true,
   "author": "Liquity AG",
-  "license": "GPL-3.0",
+  "license": "BSL-1.1",
```
danielattilasimon, line 5: Same.

bpierre (approved):

danielattilasimon (approved):
👍 👍

cvalkan (changes requested):
We should explicitly license the subgraph under MIT as we do with the frontend code.
Apart from that, the PR looks good to me.

bingen (approved):
Looks good to me.

PR #683, opened 2025-01-08
Add tests to increase coverage

bingen:


.github/workflows/contracts-tests.yml
```diff
@@ -181,10 +181,13 @@ jobs:
           lcov --remove lcov_merged.info -o lcov_merged.info
             'test/*'
             'script/*'
+            'src/Dependencies/Ownable.sol'
+            'src/Zappers/Modules/Exchanges/UniswapV3/UniPriceConverter.sol'
             'src/Interfaces/*'
```
danielattilasimon, line 186: This is not a result of the PR, but `lcov` is complaining that (and failing because) `'src/Interfaces/*'` is an unused pattern. Should we just remove it?

danielattilasimon: https://github.com/liquity/bold/actions/runs/12676876780/job/35330995484?pr=683#step:8:11

bingen: Weird. Yes, I’ll remove it

PR #687, opened 2025-01-10
Subgraph: address issues pointed out by review

bpierre:
This PR addresses some issues pointed out by @danielattilasimon’s review.

- [x] Consistent `status` updates (fix troves being stuck in redeemed mode)
- [x] Simplify `getLeverageUpdate()`
- [x] `enterBatch()` and `leaveBatch()` now return the Trove
- [x] NFT transfers support
- [x] ~Fix collateral calculations~ removed, these were not used

### DIFF
--- a/subgraph/src/TroveNFT.mapping.ts
+++ b/subgraph/src/TroveNFT.mapping.ts
@@ -1,4 +1,4 @@
-import { Address } from "@graphprotocol/graph-ts";
+import { Address, BigInt, dataSource } from "@graphprotocol/graph-ts";
 import { BorrowerInfo, Trove } from "../generated/schema";
 import { Transfer as TransferEvent } from "../generated/templates/TroveNFT/TroveNFT";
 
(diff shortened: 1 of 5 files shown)


subgraph/src/TroveNFT.mapping.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  let collId = dataSource.context().getString("collId");
+  let troveFullId = collId + ":" + event.params.tokenId.toHexString();
+
+  // update BorrowerInfo for the previous owner
+  updateBorrowerInfo(event.params.from, troveFullId, BorrowerInfoUpdate.remove);
+
+  // update BorrowerInfo for the new owner (including zero address)
+  updateBorrowerInfo(event.params.to, troveFullId, BorrowerInfoUpdate.add);
```
danielattilasimon, line 60: So burnt (closed) Troves will accumulate under the BorrowerInfo of address zero, is that right? Do we use this for something?

danielattilasimon (approved):
Looks good!

subgraph/src/TroveNFT.mapping.ts
bpierre: No I just did this for consistency, so that all debt can be queried. It’s not used at the moment.

danielattilasimon: But all Troves under address zero would have zero debt, no?

bpierre: My assumption is that the NFT of a trove could get transferred to the zero address without closing the trove. Is it incorrect?

danielattilasimon: Ah, I missed this. To answer: no, you're not allowed to transfer to the zero address:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/bd325d56b4c62c9c5c1aff048c37c6bb18ac0290/contracts/token/ERC721/ERC721.sol#L335

bpierre: Oh I didn’t know, thanks! I’ll add a check for `0x0` then :+1:

PR #696, opened 2025-01-15
test: verify V1 staking in E2E

danielattilasimon:
Found and fixed a bug in the deployment script that resulted in not being able to claim LUSD rewards from V1 staking.


contracts/script/DeployOnlyGovernance.s.sol
```diff
@@ -1,38 +0,0 @@
...
-            deployer = vm.addr(privateKey);
-            vm.startBroadcast(privateKey);
-        }
-
-        console2.log(deployer, "deployer");
-        console2.log(deployer.balance, "deployer balance");
-
-        deployGovernance(deployer, SALT, boldToken, curveFactoryAddress, usdcCurvePoolAddress);
-    }
-}
```
danielattilasimon, line 38: Should no longer be needed.

danielattilasimon:
Merging this so we can move on with the deployment script.

PR #698, opened 2025-01-15
test: Fix wrong exchange type used in Zapper test

bingen:


danielattilasimon (approved):
Looks good! Let's merge as soon as we can.

PR #707, opened 2025-01-18
[deploy-cli] fix: not being able to override env bools

danielattilasimon:
Previously, there was no way to turn on any bool options through environment variables, because we were using nullish-coalescing assignment on l-values that were never nullish (they were guaranteed to be booleans by minimist).

This was fixed by #704, but in the process, the ability to override an env "true" via an option (e.g. `--debug=false`) was lost.

This PR addresses that.


bingen (approved):
Awesome, thanks!

PR #710, opened 2025-01-18
test: Increase coverage for TroveManager

bingen:
Avoid undeterministic coverage due to invariant tests


danielattilasimon (approved):
Looks good 👍 👍

PR #711, opened 2025-01-18
fix: remove curve initiative from deployment-manifest-to-app-env.ts

chriswessels:
this util is now failing again after https://github.com/liquity/bold/commit/191e6ceafe966c051a268360bd730d03bc832e46. this pr fixes it.


danielattilasimon (approved):
Looks good to me 👍 👍 

From what I can tell this wasn't used anywhere. My understanding is that we'll get the initiatives from a separate config file, can you confirm @bpierre?

PR #712, opened 2025-01-19
fix: subgraph eventHandler type sigs

hopeyen:
fix on top of https://github.com/liquity/bold/pull/711


danielattilasimon (approved):
Thanks! Looks good!

This was changed in https://github.com/liquity/V2-gov/pull/126

PR #721, opened 2025-01-20
chore: Add Sepolia deployment

bingen:


bpierre (approved):
:green_circle::green_circle::green_circle:

I added a commit to update the app default `.env`.

PR #723, opened 2025-01-20
Deployment script minor tweaks

bingen:


danielattilasimon (approved):
Great tweaks! Especially hiding private keys from the output.

bingen:
Thanks for your additions!

PR #736, opened 2025-01-21
App: calculate hints and use them to open troves

bpierre:


frontend/app/src/liquity-utils.ts
```diff
@@ -425,3 +428,52 @@ export function usePredictAdjustInterestRateUpfrontFee(
...
+    functionName: "getSize",
+  });
+
+  const [approxHint] = await readContract(wagmiConfig, {
+    ...contracts.HintHelpers,
+    functionName: "getApproxHint",
+    args: [
+      BigInt(collIndex),
+      interestRate,
+      15n * numTroves, // (15 * troves) gives a hint close to the right position
```
bingen, line 463: This does what the description in the README says, but the original documentation in the function header says it should be `15*sqrt(numTroves)`.
I have some comments about it:
- I don’t remember where the magic number 15 comes from
- Using `sqrt(numTroves)` makes a lot of sense, because then we split the complexity: `O(n^(1/2))` for each of the functions, `getApproxHint` and `findInsertPosition`. But I guess it may happen that it’s still too much if there are too many troves. I don’t remember if we estimate when this would happen, but anyway I think we are good for now, we can fix it later when we approach such a high number.

bingen (approved):
Looks good to me

frontend/app/src/liquity-utils.ts
danielattilasimon: Agree, we should use some multiple of `sqrt(numTroves)` and consider fixing the code in the README.

We do say this a few paragraphs above:
> It can be shown mathematically that for `numTrials = k * sqrt(n)`, the function's gas cost is with very high probability worst case `O(sqrt(n))` if `k >= 10`.

In my experience, `k` doesn't need to be that high, even `k = 1` works well in practice.

danielattilasimon (approved):
Works great!

PR #738, opened 2025-01-21
Improve INSTRUCTIONS.md (production build instructions)

bpierre:


INSTRUCTIONS.md
```diff
@@ -1,48 +1,111 @@
...
+cp .env .env.local
+```
+
+Edit the newly created `.env.local` to set the app [environment variables](./frontend/app/README.md#environment).
+
+Optional: you can generate a set of contract addresses for the `.env.local` file by running the following command:
+
+```bash
+# Replace the address file with the one you want to use
+cd ../contracts
```
ColinPlatt, line 50: assume that we're in frontend/apps

```bash
cd ../../contracts
```

INSTRUCTIONS.md
```diff
@@ -1,48 +1,111 @@
...
+pnpm tsx ./utils/deployment-manifest-to-app-env.ts ./addresses/11155111.json
```
ColinPlatt, line 51: should the manifest script just be copied in or should we modify it to replace the destination file?

should we add the instructions for the destination file in the command?

INSTRUCTIONS.md
```diff
@@ -1,48 +1,111 @@
...
+```
+
+This is useful if you want to use a different set of contracts than the ones provided in the default `.env` file.
+
+You can now build the app for production:
```
ColinPlatt, line 56: from frontend/app

ColinPlatt (changes requested):
can you check the deployment-manifest-to-app-env.ts is in line with main?

INSTRUCTIONS.md
```diff
@@ -1,48 +1,111 @@
...
+
+```bash
+pnpm build-deps # only needed the first time
+pnpm build
+```
+
+The app will be built in the `out/` directory which can be deployed to any static hosting service.
+
+## Local development setup
```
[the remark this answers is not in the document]
bpierre: Local deployment, which might be a fork or not. It’s not great, so maybe we can remove it for now for clarity. I think it would be nice to unify all the documentation in this file (contracts, subgraph, frontend/app, frontend/uikit).

INSTRUCTIONS.md
bpierre: The manifest is only read by the script, which then outputs the env vars so users can paste it (or redirect the command to the `.env.local` with `>>` on macOS / Linux)

INSTRUCTIONS.md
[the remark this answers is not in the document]
ColinPlatt: ok.

can leave for now, but let's add instructions to `pnpm dev` against testnet/mainnet later stp

INSTRUCTIONS.md
ColinPlatt: understood, would add that to the instructions here.

afaik there is an `--append` flag that achieves the same as `>>`

bpierre: Ah yes true! Do you think it’s important to let users know about this? My thinking was that just outputting by default and letting users copy and paste is good, so people can verify what they are doing. Advanced users can check the script `--help` or use `>>`.

PR #740, opened 2025-01-22
feat: use less-than sign for tiny numbers

### LINKED ISSUE #763: Staking rewards round up in a misleading way
My actual ETH rewards, as per the tooltip are 0.000006... ETH, so perhaps this should show `<0.01 ETH`

Image

danielattilasimon:
Showing just 0.01 for numbers less than 0.01 but greater than zero can be misleading, as one might think that the number rounds to 0.01, which may not be the case e.g. in the case of tiny ETH rewards from V1 staking.

Fixes #763.


ColinPlatt:
Will we add this or has this been replaced with another PR?

bpierre:
@ColinPlatt we can continue from here, I am planning to add a commit to this PR today

PR #746, opened 2025-01-22
Update the app env for mainnet

bpierre:


frontend/app/.env
```diff
@@ -12,86 +12,142 @@ NEXT_PUBLIC_COINGECKO_API_KEY=
...
+######################
+
+NEXT_PUBLIC_CHAIN_BLOCK_EXPLORER=Etherscan|https://etherscan.io
+NEXT_PUBLIC_CHAIN_CONTRACT_ENS_REGISTRY=0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e
+NEXT_PUBLIC_CHAIN_CONTRACT_ENS_RESOLVER=0xce01f8eee7E479C928F8919abD53E553a36CeF67|19258213
+NEXT_PUBLIC_CHAIN_CONTRACT_MULTICALL=0xca11bde05977b3631167028862be2a173976ca11
+NEXT_PUBLIC_CHAIN_CURRENCY=Ether|ETH|18
+NEXT_PUBLIC_CHAIN_ID=1
+NEXT_PUBLIC_CHAIN_NAME=Ethereum
+NEXT_PUBLIC_CHAIN_RPC_URL=https://cloudflare-eth.com
```
bpierre, line 26: @danielattilasimon @ColinPlatt let me know if you think we should use another one

danielattilasimon: IMO a public RPC is a fine default, and it should be left to frontend operators to get a paid API key if they so choose.

danielattilasimon (approved):
Looks good to me! Checked the addresses again, all look good!

PR #759, opened 2025-01-23
doc: Minor changes to frontend local instructions

bingen:


INSTRUCTIONS.md
```diff
@@ -46,9 +46,9 @@ Edit the newly created `.env.local` to set the app [environment variables](./fro
 Optional: you can generate a set of contract addresses for the `.env.local` file by running the following command:
 
 ```bash
-cd contracts
+cd ../../contracts
 # Replace the address file with the one you want to use
-pnpm tsx ./utils/deployment-manifest-to-app-env.ts ./addresses/11155111.json
+pnpm tsx ./utils/deployment-manifest-to-app-env.ts ./addresses/11155111.json > ../frontend/app/.env.local
```
bpierre, line 51: This would override the other options in the `.env.local`. We could append, but I was thinking that just outputting by default should be enough to avoid any manipulation error? People that know how redirections work can just use them.

INSTRUCTIONS.md
```diff
@@ -46,9 +46,9 @@ Edit the newly created `.env.local` to set the app [environment variables](./fro
 Optional: you can generate a set of contract addresses for the `.env.local` file by running the following command:
 
 ```bash
-cd contracts
+cd ../../contracts
```
bpierre, line 49: I made all the `cd` instructions relative to the root directory, we tried using relative paths (https://github.com/liquity/bold/pull/738) but some steps are optional which makes it difficult to follow. Maybe we could use something like `ROOT/contracts` for all paths to make it super explicit?

PR #792, opened 2025-01-29
Earn: fix APRs

### LINKED ISSUE #775: [UI] Add historical APR figures
Current APR figures are improperly labelled and only show borrow income less upfront fees, these figures should both be available.

bpierre:
fixes #775
requires https://github.com/liquity/api.liquity.org/pull/748

Earn pools APRs are now fetched from the Liquity stats.


ColinPlatt:
Have changed content in the homescreen and aligned the column headers

PR #826, opened 2025-02-08
[WIP] Show how approx dx can be quoted

danielattilasimon:


ColinPlatt:
is this still needed or can it be closed?

danielattilasimon:
> is this still needed or can it be closed?

It will be needed for the leverage feature, so let's keep it for now.

PR #838, opened 2025-02-18
fix(nft-metadata-import): fixing the absolute import in MetadataNFT.sol

SocksNFlops:
## Changes:
- Removing the absolute import in `MetadataNFT.sol` since it breaks downstream remappings
- Attempting to import this repo as a forge dependency creates a conflicting remapping and throws this error:
```
>>> ✗ forge build
[⠊] Compiling...
[⠃] Compiling 206 files with Solc 0.8.24
[⠊] Solc 0.8.24 finished in 795.52ms
Error: Compiler run failed:
Error (9553): Invalid type for argument in function call. Invalid implicit conversion from enum ITroveManager.Status to enum ITroveManager.Status requested.
  --> lib/bold/contracts/src/TroveNFT.sol:44:22:
   |
44 |             _status: troveManager.getTroveStatus(_tokenId)
```
This is fixed by removing the only absolute import in the entire codebase (probably a typo) and replacing it with a relative one.

This does not change any functionality or structure since the imports are still the same files. It simply fixes the conflicting remappings.


bingen (approved):
Looks good. Thanks!

Issue #839, opened 2025-02-18
[Contract] Absolute Import in MetadataNFT breaks downstream remappings


SocksNFlops:
## Issue:
`MetadataNFT.sol` has an absolute import on line 12:
```
import {ITroveManager} from "../Interfaces/ITroveManager.sol";
```
This creates issues downstream when using this repo as a dependency in forge. Namely, if you use a remapping such as this one:
```
@bold=lib/bold/contracts/src/
```
It will not be able to resolve the multiple references of `ITroveManager`  due to the fact that `src` is changed in the base path.

## Error:
The following error is thrown whenever `forge build` is ran downstream:
```
>>>>>> ✗ forge build
[⠊] Compiling...
[⠃] Compiling 206 files with Solc 0.8.24
[⠊] Solc 0.8.24 finished in 795.52ms
Error: Compiler run failed:
Error (9553): Invalid type for argument in function call. Invalid implicit conversion from enum ITroveManager.Status to enum ITroveManager.Status requested.
  --> lib/bold/contracts/src/TroveNFT.sol:44:22:
   |
44 |             _status: troveManager.getTroveStatus(_tokenId)
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Fix:
This is a 3-character fix: https://github.com/liquity/bold/pull/838
Consists of just changing the absolute import for a relative import

TristanHyped:
+ 1. please fix LQTY team!

SocksNFlops:
In case my fix isn't merged, this work-around with remappings worked for me:
```
lib/bold:lib/Solady/=lib/bold/contracts/lib/Solady/
lib/bold:src/Interfaces/=lib/bold/contracts/src/Interfaces/
```

There are two cases of absolute imports that will break the downstream and this works in overriding both of those problems.

PR #851, opened 2025-02-25
fix: Change import path for Solady dependency

bingen:
See: https://github.com/liquity/bold/issues/839#issuecomment-2673019668


danielattilasimon (approved):
👍 👍

PR #863, opened 2025-03-20
fix: Remove unused var from struct

### LINKED ISSUE #862: unused tmp variable in BorrowerOperations
https://github.com/liquity/bold/blob/b2ff26c5f09e72eaa5ad7eb24210aded4e80f00e/contracts/src/BorrowerOperations.sol#L118

this variable seems to not be used anywhere (found it when hunting for a few bytes to save in contract size)

assuming this is just an innocent oversight and can safely be deleted, but you never know...

bingen:
Fixes #862


ColinPlatt:
send it!

danielattilasimon (approved):
👍 👍

RickGriff (approved):
Looks good!

PR #888, opened 2025-04-29
Liquidation grace period

bingen:


danielattilasimon (approved):
Looks good! ⚡ ⚡

contracts/src/TroveManager.sol
```diff
@@ -464,6 +464,18 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         return _status == Status.active || _status == Status.zombie;
     }
 
+    // Troves with debt adjusted in the same block cannot be liquidated
+    // This is to prevent attacks involving sandwiching a price update to self-liquidate
+    function _isRecent(uint256 _troveId) internal view returns (bool) {
+        if (uint256(Troves[_troveId].lastDebtUpdateTime) >= block.timestamp) return true;
+        address batchAddress = _getBatchManager(_troveId);
+        if (batchAddress != address(0) && uint256(batches[batchAddress].lastDebtUpdateTime) >= block.timestamp) {
```
RickGriff, line 472: This is interesting. It means that _any_ borrower in a batch can give _all_ Troves in that batch a 1-block liquidation delay, right?
  
But doesn't that mean they could save other Troves from liquidation even though they're below MCR?

e.g. Troves A, B, C in batch, with :

CR_A: 109%
CR_B: 109%
CR_C: 120%

C could slightly adjust his debt every block and save B and A from liquidation, right?

With non-batched Troves it doesnt matter since any adjustment must bring the Trove above MCR, so it really is a liquidation grace period. But with batched Troves, it seems like C can grief liquidation of A and B indefinitely, even though they're below MCR.

contracts/src/TroveManager.sol
```diff
@@ -464,6 +464,18 @@ contract TroveManager is LiquityBase, ITroveManager, ITroveEvents {
         return _status == Status.active || _status == Status.zombie;
     }
 
+    // Troves with debt adjusted in the same block cannot be liquidated
+    // This is to prevent attacks involving sandwiching a price update to self-liquidate
+    function _isRecent(uint256 _troveId) internal view returns (bool) {
+        if (uint256(Troves[_troveId].lastDebtUpdateTime) >= block.timestamp) return true;
```
RickGriff, line 470: An individual Trove's `lastDebtUpdateTime` is updated when they open or adjust debt/coll but also when they adjust their interest rate.

IIRC, interest rate adjustments don't actually enforce that the resulting ICR >= MCR (or do they - am I missing that?) So what if a Trove at ICR < MCR keeps adjusting interest rate every block - can they avoid liquidation indefinitely?

RickGriff (commented):
A couple of questions re:griefing here. Thanks!

RickGriff (commented):
I think we could solve both the individual and batch issues I mentioned - by using another separate timestamp. 

Something like `lastTouchTime` - that _only_ gets updated when the borrower touches the coll/debt of their own Trove. That way:

- rate adjustments on individual Troves with ICR < MCR don't update it, and so cant grief liquidations
- batch adjustments on other Troves in the same batch don't affect the timestamp for the Trove, and so can't grief liquidations of it

contracts/src/TroveManager.sol
bingen: I would say we should enforce MCR on adjusting interest. Not only for this, but also because it helps system stability.
The more we force troves under MCR to go back above it the better in my opinion.

contracts/src/TroveManager.sol
bingen: Good point!
Not sure how to proceed, but I’m leaning towards dismissing the PR.
We need to make sure we block all possible self-triggered liquidations, otherwise it’s not worth making changes.
- If we don’t check last batch debt update, there would be a hole there.
- If we leave it as is, in normal conditions it wouldn’t be an issue, as having to touch trove C (in your example), and beat all liquidation bots to make sure the touch goes first seems unpractical, expensive and unrealistic. But obviously it opens a door for griefing "theoretical" attacks, where you are able to maintain a huge trove (A or B) below MCR, until it could even provoke a shutdown.

contracts/src/TroveManager.sol
bingen: It is actually enforced, in `_applyUpfrontFee` function.

bingen: See:
https://github.com/liquity/bold/blob/6e61065a3cbe7939f7f170614376774bba50df5e/contracts/src/BorrowerOperations.sol#L535
https://github.com/liquity/bold/blob/6e61065a3cbe7939f7f170614376774bba50df5e/contracts/src/BorrowerOperations.sol#L1134

RickGriff: Ah cool, thanks.  That would stop indefinite individual griefing at least - adjusting every block would incur premature fees and thus this check then.

bingen:
Close because of the problem with batches mentioned above.

PR #889, opened 2025-04-29
Remove 0.5% coll gas compensation from Redistributions

bingen:


danielattilasimon (approved):
Looks good! 👍 👍

contracts/test/TestContracts/InvariantsTestHandler.t.sol
```diff
@@ -2649,33 +2649,35 @@ contract InvariantsTestHandler is Assertions, BaseHandler, BaseMultiCollateralTe
...
         uint256 collSPPortion = collRemaining * spOffset / trove.entireDebt;
-        uint256 spCollGain = Math.min(collSPPortion, spOffset * (_100pct + LIQ_PENALTY_SP[i]) / _price[i]);
+
+        // Deduct coll gas comp from SP portion
+        uint256 collGasComp = Math.min(collSPPortion / COLL_GAS_COMPENSATION_DIVISOR, COLL_GAS_COMPENSATION_CAP);
+        t.collGasComp += collGasComp;
+        collRemaining -= collGasComp;
+
+        // Send remainder of SP portion to SP, capped by liq penalty
+        uint256 spCollGain = Math.min(collSPPortion - collGasComp, spOffset * (_100pct + LIQ_PENALTY_SP[i]) / _price[i]);
```
RickGriff, line 2667: Are the units right here? `spOffset` is in units of BOLD, whereas the first term is in units of collateral? Do we need to scale the second down by the price?

contracts/test/TestContracts/TroveManagerTester.t.sol
```diff
@@ -103,6 +103,17 @@ contract TroveManagerTester is ITroveManagerTester, TroveManager {
         return LiquityMath._computeCR(_coll, _debt, _price);
     }
 
+    function getCollGasCompensation(uint256 _entireColl, uint256 _entireDebt, uint256 _boldInSPForOffsets)
+        external
+        pure
+        returns (uint256)
+    {
+        if (_boldInSPForOffsets < _entireDebt) {
+            _entireColl = _entireColl * _boldInSPForOffsets / _entireDebt;
```
RickGriff, line 112: Nitpick: not technically `_entireColl` any more after this op, right? (but the result passed down is correct)

RickGriff (approved):
Looks good!

contracts/test/TestContracts/InvariantsTestHandler.t.sol
danielattilasimon: It _is_ scaled by the price, isn't it?
```solidity
spOffset * (_100pct + LIQ_PENALTY_SP[i]) / _price[i]
```

RickGriff: ...yes! Sorry, somehow missed that

PR #890, opened 2025-04-29
fix: RemoveManager assumes AddManager

bingen:


contracts/src/BorrowerOperations.sol
```diff
@@ -571,13 +571,12 @@ contract BorrowerOperations is LiquityBase, AddRemoveManagers, IBorrowerOperatio
...
         address owner = troveNFT.ownerOf(_troveId);
-        address receiver = owner; // If it’s a withdrawal, and manager has receive privilege, manager would be the receiver
+        address receiver = owner; // If it’s a withdrawal, and remove manager privilege is set, a different receiver can be defined
 
         if (_troveChange.collDecrease > 0 || _troveChange.debtIncrease > 0) {
             receiver = _requireSenderIsOwnerOrRemoveManagerAndGetReceiver(_troveId, owner);
-        }
-
-        if (_troveChange.collIncrease > 0 || _troveChange.debtDecrease > 0) {
+        } else if (_troveChange.collIncrease > 0 || _troveChange.debtDecrease > 0) {
```
danielattilasimon, line 578: Technically, in all cases where we _don't_ enter the `else if`, we revert later with `_requireNonZeroAdjustment`, so we could simply do an `else`, especially if we move up `_requireNonZeroAdjustment`.

danielattilasimon (approved):
Looks good to me! 👍 👍

danielattilasimon:
We should update the REAME accordingly.

contracts/src/BorrowerOperations.sol
bingen: Good point!

bingen: And actually, in any case it would be weird/unsafe that we allow to call `_adjustTrove` by an address which is not owner nor remove manager nor add manager.

RickGriff (approved):
Looks good!

PR #893, opened 2025-05-01
fix: Require Zapper to be the receiver for Remove operations

bingen:


danielattilasimon (approved):
Looks good! 👍 👍

contracts/test/zapperGasComp.t.sol
```diff
@@ -405,6 +473,44 @@ contract ZapperGasCompTest is DevTestSetup {
...
+            removeManager: address(0),
+            receiver: address(0)
+        });
+        vm.startPrank(A);
+        uint256 troveId = gasCompZapper.openTroveWithRawETH{value: ETH_GAS_COMPENSATION}(params);
+        vm.stopPrank();
+
+        vm.startPrank(A);
+        // Add a remove manager for the zapper
+        gasCompZapper.setRemoveManagerWithReceiver(troveId, B, A);
```
RickGriff, line 502: This call doesn't actually set the receiver in the core, right (only in the zapper)?

Receiver in core is only set in the next line AFAICS. No big deal though - was just a little confused by "Change receiver"

RickGriff (approved):
Looks good!

contracts/test/zapperGasComp.t.sol
bingen: Yes, that’s correct, but it’s still "change", because the receiver in core is set on opening trove.

PR #899, opened 2025-05-05
fix: Add slippage protection to close trove from collateral in Zappers

bingen:


contracts/src/Zappers/GasCompZapper.sol
```diff
@@ -275,6 +281,8 @@ contract GasCompZapper is BaseZapper {
         require(msg.sender == address(flashLoanProvider), "GCZ: Caller not FlashLoan provider");
 
         LatestTroveData memory trove = troveManager.getLatestTroveData(_params.troveId);
+        uint256 collLeft = trove.entireColl - _params.flashLoanAmount;
```
RickGriff, line 284: In the WETHZapper we account for `ETH_GAS_COMPENSATION` too. Is intentionally left out here (if, so why)?

RickGriff (approved):
Looks good!

contracts/src/Zappers/GasCompZapper.sol
danielattilasimon: In case of WETH, the collateral and gas compensation tokens are the same, so we can send it to the user in one combined transfer (of raw ETH, following a withdrawal from WETH), hence the addition of `ETH_GAS_COMPENSATION`.

danielattilasimon: Of course, we could have kept `ETH_GAS_COMPENSATION` out of `collLeft` and dealt with it later, when doing the withdrawal & transfer.

The downside of the current implementation is that the frontend has to calculate `minExpectedCollateral` differently between (W)ETH and other collateral branches.

bingen: Yes, because gas compensation is WETH, while here coll will be another token (wstETH or RETH).

bingen:
I see you both have concerns about how I’m dealing with gas compensation. I did like it is on purpose, so intuitively what you have to pass to the function is what you expect to receive as transfer at the end.
It’s true that the behaviour is different for each zapper, but that’s because they are inherently different. Besides, this is something mostly the frontend needs to know. I assume most users wouldn’t care about this.
But happy to change it if you think we should do it differently.

danielattilasimon:
> I did like it is on purpose, so intuitively what you have to pass to the function is what you expect to receive as transfer at the end.

IMO it's inconsistent. In the case of WETH, you have to specify that you also expect to receive 0.0375 ETH gas compensation. This is not the case for any other collateral type, even though the gas compensation still applies.

Also, the parameter says `minExpectedCollateral`, but the gas compensation deposit is not collateral, it just happens to be denominated in the same currency as collateral in the case of WETH.

PR #919, opened 2025-05-09
fix: Comment about P range in SP

bingen:


RickGriff (approved):
Yes, this looks right, since, inside the `while` loop:
- As long as `newP` evaluates to `<1e27` we multiply it by `1e9`. 
- When initial `newP % 1e9 != 0`, then final `newP > 1e27'
- When  initial`newP % 1e9 == 0`, then final `newP == 1e27`

And it can't become larger than `1e36`, since we only scale it up by `1e9` when it's below `1e27`.  So it can only be `1e36` before the first offset.

So the bounds on `P` are `[1e27, 1e36]`.

Then as you say, the floor division can only evaluate to 0 when `totalBoldDeposits > P * (totalBoldDeposits - _debtToOffset)`,  and with the lower bound on `P` and the `MIN_BOLD_IN_SP`, that's `totalBoldDeposits > 1e27 * 1e18`

danielattilasimon (commented):
Looks good to me to! 👍 👍

Issue #1011, opened 2025-07-04
[App] Hide new iniatives that are still in warm-up mode


UncleBlock:
Background

- Users can vote on new initiatives once the warm up period has elapsed (next epoch?)

Requirement

- The frontend should only display initiatives that can be up/down voted
- Initiatives that are not active yet should be hidden (same for deactivated initiatives but this should already work)

PR #1015, opened 2025-07-07
App: Hide new initiatives that are still in warm-up mode

cyril-dfi:
Solves https://github.com/liquity/bold/issues/1011


danielattilasimon (approved):
LGTM! ⚡ ⚡

PR #1059, opened 2025-07-25
[App] Exclude user's own debt from debt-in-front calculation

### LINKED ISSUE #994: One's own debt should be excluded from debt-in-front calculation
When adjusting the interest rate of an existing position, its debt will be included in the original interest rate bracket when calculating debt-in-front for the new interest rate. For large positions, this can introduce a significant error in the debt-in-front figure.

cyril-dfi:
fixes #994


danielattilasimon (approved):
Looks good 👍 👍

Issue #1055, opened 2025-07-25
[App] Loan card layout too cramped when adjusting


danielattilasimon:
In the case of wstETH, even a 2 digit collateral causes clipping.

<img width="551" height="248" alt="Image" src="https://github.com/user-attachments/assets/ba867a01-fdda-44f4-9580-aab8f9c3e090" />

Instead of 3x2, we might want to consider a 2x3 layout. Mockup of how that might look like:

<img width="551" height="307" alt="Image" src="https://github.com/user-attachments/assets/b0dadf0c-d95a-43af-8bbd-ef2af0c12e88" />

UncleBlock:
@danielattilasimon I just realized that this is the tile of the detail view of a position and not the tile on the dashboard itself. I was mainly worried about the inconsistency between tiles on the dash.
So I think your suggestion is fine.

PR #1095, opened 2025-08-13
[App] Change loan card layout to 2 columns and 3 rows to avoid clipping

hutsul:
Fixes https://github.com/liquity/bold/issues/1055


frontend/app/src/screens/TransactionsScreen/LoanCard/components/ClosedLoan/components/CollateralRow/CollateralRow.tsx
```diff
@@ -0,0 +1,33 @@
+import { GridItemWrapper } from "@/src/screens/TransactionsScreen/LoanCard/components/components/GridItemWrapper";
```
danielattilasimon, line 1: Why is there a `components` directory inside another `components` directory?

danielattilasimon:
After your changes, my loan is incorrectly shown as a leverage position instead of a loan. Before:
image

After:
image

Issue #1035, opened 2025-07-14
[App] Improve Redemption Risk Calculation (3-5d)


UncleBlock:
Background

- The risk shown on the frontend varies for the branches although the same rate setting algorithms are used (see screenshot from ICP rate manager which is aiming at 50% debt in front => conservative)
- ~~Can it be that for the calculation of the risk the weighting of the branches is not considered~~
- ~~Could be related to issue #994~~

Goal

- The redemption risk calculations is consistent across branches
- Ideally, the risk indicators are aligned with the ICP and other delegate risk framework (or we can document why we consider sth. low, medium or high risk on the frontend)

<img width="1130" height="896" alt="Image" src="https://github.com/user-attachments/assets/a4acb01e-9f81-498c-a700-f166001a092f" />
<img width="2352" height="952" alt="Image" src="https://github.com/user-attachments/assets/a6775ef2-ed9e-479c-a6d5-5964bbe40204" />

danielattilasimon:
This is a limitation of the way we currently calculate debt-in-front, which is a conservative estimation based on interest rate brackets of 0.1% granularity.

In your screenshot, the redemption risks of the rETH and ETH positions are overestimated, because in both cases there are very large positions within the same brackets as the positions themselves, presumably at slightly lower interest rates than those set by the ARM. Since we currently don't know the precise distribution of debt within a bracket, we can't tell the amount of debt in front of each position within their own bracket, so we conservatively assume 0 BOLD.

One way to improve the calculation would be to use a read-only helper contract that calculates the exact amount of debt in front of the chosen position within the bracket. To facilitate this, we'd also have add extra data to the subgraph in the form of one arbitrary Trove ID per inhabited bracket, to serve as hints for the iteration.

UncleBlock:
My thoughts:

- I think debt in front is an important data point and it would be good to make it more reliable (as it otherwise results in user frustrations aka redemptions)
- On the other hand, I don't like to introduce more The Graph dependency and complexity
- Could we do it in a way that we improve it but still keep a fall back in case the additional The Graph information is not available (or are we anyway already heavily dependent and it doesn't really make a difference)?

UncleBlock:
See input / discussion on Slack about debt in front issue on the frontend: https://liquity.slack.com/archives/C076ESVS67M/p1753082628096699

PR #1108, opened 2025-08-21
[Contracts] Debt-in-front helper

danielattilasimon:
Needed for #1035


danielattilasimon:
It's not used in the frontend yet, but it'll be used exactly as you described.

PR #1126, opened 2025-09-15
test: Add mainnet tests for UniV4MerklRewards

bingen:


contracts/test/InitiativeUniV4Merkl.t.sol
```diff
@@ -0,0 +1,115 @@
...
+        assertEq(campaign.rewardToken, params.rewardToken, "rewardToken");
+        assertEq(campaign.amount, params.amount * 97 / 100, "amount minus fees");
+        assertEq(campaign.campaignType, params.campaignType, "campaignType");
+        assertEq(campaign.startTimestamp, params.startTimestamp, "startTimestamp");
+        assertEq(campaign.duration, params.duration, "duration");
+        assertEq(campaign.campaignData, params.campaignData, "campaignData");
+
+        assertGt(
+            boldToken.balanceOf(address(merklDistributionCreator.distributor())),
+            0,
```
RickGriff, line 110: If the campaign has been immediately created, why should this be > 0? Does the campaign pull some (or all) funds immediately from initiative upon creation?

RickGriff (approved):
Looks good!

contracts/test/InitiativeUniV4Merkl.t.sol
danielattilasimon, line 110: Earlier, we assert that the entire BOLD is gone from `uniV4MerklRewardsInitiative`. Is this where it goes (minus fees)? If so, can we assert exactly how much should be here?

danielattilasimon (approved):
Looks good 👍 👍

contracts/test/InitiativeUniV4Merkl.t.sol
```diff
@@ -0,0 +1,115 @@
...
+        governance.registerInitiative(address(uniV4MerklRewardsInitiative));
+        vm.stopPrank();
+
+        assertGt(governance.registeredInitiatives(address(uniV4MerklRewardsInitiative)), 0, "Initiative should be registered");
+
+        skip(7 days);
+
+        // Allocate to initiative
+        _allocateLQTY_begin(staker);
+        _allocateLQTY_vote(address(uniV4MerklRewardsInitiative), int256(lqtyStake)); // TODO
```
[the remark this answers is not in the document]
bingen: Nothing that I remember, so I’ll remove the comment. Thanks!

contracts/test/InitiativeUniV4Merkl.t.sol
bingen: Yes, exactly:
https://github.com/AngleProtocol/merkl-contracts/blob/c621948937be3dac367e0376601f93201adefb3a/contracts/DistributionCreator.sol#L510

bingen: Yes, fixing it.

Issue #1107, opened 2025-08-20
[App] Improve error messages for manual delegation strategy selection


UncleBlock:
Issue

- Users might copy an strategy but not notice that it is not suitable for their Trove collateral

Solution / Requirement

- Distinguish between these two error messages
- 1) Error message 1: "This is not a valid delegate address".
- 2) Error message 2: "This delegate does not support your Trove collateral. Choose a delegate for you collateral."
- If we can easily fetch the Trove collateral from the user we could include it in the message above: "Choose a delegate for you collateral (wstETH)."

Current error message: This is a valid delegate but not for this collateral type: 
<img width="622" height="349" alt="Image" src="https://github.com/user-attachments/assets/f52fef7a-4979-4bbc-9093-1e9716713e9d" />

PR #1127, opened 2025-09-15
[App] Delegates metadata

### LINKED ISSUE #970: [App] Allow for logging of delegates with look up for metadata
should allow for name and description for known-delegates (pass via API?)

cyril-dfi:
Closes #970

### DIFF
--- a/frontend/app/src/liquity-delegate.ts
+++ b/frontend/app/src/liquity-delegate.ts
@@ -38,7 +38,7 @@ export function useDelegateDisplayName(delegateAddress: string | null) {
   const knownDelegatesQuery = useKnownDelegates();
 
   return useMemo(() => {
-    if (!delegateAddress || !knownDelegatesQuery.data) return null;
+    if (!delegateAddress || !knownDelegatesQuery.data) return undefined;
 
     // branchId could be used for filtering delegates by collateral in the future
     for (const group of knownDelegatesQuery.data) {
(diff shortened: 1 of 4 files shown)


cyril-dfi:
If this is released as is, we'll need to update this blog post with the new screenshots/methodology: https://www.liquity.org/blog/autonomous-interest-manager

frontend/app/src/delegates.json
danielattilasimon, line 1: Instead of baking this file into the frontend, we should fetch it from a configurable URL `NEXT_PUBLIC_KNOWN_DELEGATES_URL`, which would default to `https://api.liquity.org/v2/known-delegates/ethereum.json`. Then we will add this file to the API repo.

The reason is we want frontend operators to be able to override the delegates (e.g. by placing a JSON in their `public` directory at build time and pointing the URL at `https://their-frontend.app/delegates.json`) but still be able to add new delegates without releasing a new app version for those frontends that choose not to override them.

cyril-dfi:
This latest commit https://github.com/liquity/bold/pull/1127/commits/942493b0d01d31c6fdac22ed23fc39d037e0ad50  closes https://github.com/liquity/bold/issues/1107

frontend/app/src/comps/InterestRateField/InterestRateField.tsx
```diff
@@ -157,8 +157,6 @@ export const InterestRateField = memo(
       )
       : debtInFrontOfInterestRate.data;
 
-    const redemptionRisk = debtInFront
-      && getRedemptionRisk(debtInFront.debtInFront, debtInFront.totalDebt);
```
danielattilasimon, line 161: Why are we no longer calculating redemption risk???

Issue #927, opened 2025-05-13
[App] Better error handling (prevent TXs that would fail)


ColinPlatt:
Background
- Frontend lets you do some txs that will fail.
- E.g. error for opening troves where TCR goes under CCR (and it doesn't even give an error message)

Solution
- Handle error for opening/adjusting troves where TCR goes under CCR (check future TCR and ensure tx will pass)
- add tooltip with explanation to UI (and link to docs)

Open
- It needs review and a proper list of cases we want to improve
- E.g. balance checks

PR #1144, opened 2025-10-03
[App] Better error handling (prevent TXs that would fail)

cyril-dfi:
Closes https://github.com/liquity/bold/issues/927


danielattilasimon (approved):
Looks great! 👍 👍

Issue #1065, opened 2025-07-31
[App] Redemption View - Add appropriate Disclosures and Warnings


UncleBlock:
Background

- Bluechip ratings take into consideration if users can easily redeem their stable on the frontend. If not the rating is lower.
- Therefore, it would be good to have a redemption possibility on the frontend for users.
- But as there are known risks with redemptions (sandwiching or see cantina issues) we should make users aware of it

Goal 

- Improve frontend so that users faces a minimal risk of losses

Frontends Requirements
 - Add Zapper to make redemptions safe (pass estimated values and have some tolerance built in)
 - Build in tolerance value instead of max (fee slipage tolerance with a good default)
 - Add also estimate rate (and color it red if it becomes really high?)
 - Adjust max redemption fee based on the BOLD amount entered
 - Deep Link to docs
 - Call it redemption „fee“ on the frontend

Summary Scree Requirements
 - Add estimate: Value of BOLD and Collateral 
 - Show exact fee and potentially also the dollar value
 - Provide warning if the estimated fee is higher than fee range

Doc Requirements

- Make the redemption fee section the last section in the redemption part
- Add additional section that gives some voice over for the redemption view (why three collateral, why the mix, slippage fee, etc.)

Current info section(see below)

<img width="1116" height="318" alt="Image" src="https://github.com/user-attachments/assets/d54633c7-ae65-47a0-9077-a2369a5bd307" />

PR #1164, opened 2025-10-22
[Contracts] Redemption helper

danielattilasimon:
Adds a helper contract that will be used by the frontend to implement #1065.

### Slippage protection

Normally, a redemption is performed by a bot in a closed arbitrage transaction, usually through a private mempool and with the help of a flashloan. Thus, if there's slippage of the redemption fee, redemption proportions or the execution price of any of the trades involved in the transaction making the redemption unprofitable, the TX will simply fail or not be included at all.

This is not the case when redeeming directly through `CollateralRegistry`: the backend will only ensure that the redemption fee (as a percentage) doesn't exceed the maximum acceptable value passed by the redeemer, it doesn't care if the collateral prices (as reported by their oracles) or the proportions of unbacked debt between branches has changed compared to what the redeemer simulated before signing their transaction.

This helper contract implements a wrapper around `redeemCollateral()` that adds the new parameter `_minCollRedeemed`, ensuring that the redeemer gets at least as much collateral from each branch as was presented to them on the UI (minus a small slippage tolerance margin, to be calculated by the frontend).

### Redemption optimization

The job of `CollateralRegistry` is to break a redemption down into multiple smaller redemptions (one per each branch) adding up to the total amount that is to be redeemed. Additionally, it is responsible for calculating the redemption fee (as a percentage). It does so at the beginning of the redemption sequence, assuming that each of the per-branch redemptions will be complete, i.e. that there will be no unredeemed BOLD left.

However, this is not always the case. For example, the redemption could terminate early on one or more branches due to hitting the user-configured iteration limit (`_maxIterationsPerCollateral`). When this happens, the user pays a higher fee than warranted by the actual amount of BOLD successfully redeemed, because of the aforementioned assumption made when calculating the fee.

The `truncateRedemption()` function lets the user avoid such (unlikely) scenarios by simulating the redemption, and calculating the maximum amount of BOLD that can be redeemed completely within the given iteration limit.

### DIFF
--- a/contracts/test/RedemptionHelper.t.sol
+++ b/contracts/test/RedemptionHelper.t.sol
@@ -15,11 +15,6 @@ import {DevTestSetup} from "./TestContracts/DevTestSetup.sol";
 uint256 constant NUM_BRANCHES = 3;
 uint256 constant NUM_TROVES = 20;
 
-function array(uint256 a, uint256 b, uint256 c) pure returns (uint256[] memory r) {
-    r = new uint256[](3);
-    (r[0], r[1], r[2]) = (a, b, c);
-}
-
 contract RedemptionHelperTest is DevTestSetup {
     using Strings for *;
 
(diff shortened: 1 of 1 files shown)


contracts/src/RedemptionHelper.sol
```diff
@@ -0,0 +1,165 @@
...
+            branch[i].collBalanceBefore = branch[i].collToken.balanceOf(address(this));
+        }
+
+        uint256 boldBalanceBefore = boldToken.balanceOf(address(this));
+
+        boldToken.transferFrom(msg.sender, address(this), _bold);
+        collateralRegistry.redeemCollateral(_bold, _maxIterationsPerCollateral, _maxFeePct);
+
+        for (uint256 i = 0; i < numBranches; ++i) {
+            branch[i].collRedeemed = branch[i].collToken.balanceOf(address(this)) - branch[i].collBalanceBefore;
```
bingen: Why do we put `collRedeemed` in the branch struct array? As we are only using it inside this loop, can’t it just be a temporary variable here?

contracts/src/RedemptionHelper.sol
```diff
@@ -0,0 +1,165 @@
...
+        (SimulationContext[] memory branch, uint256 totalProportions) =
+            simulateRedemption(_bold, _maxIterationsPerCollateral);
+
+        if (totalProportions == 0) return (0, 0, redeemed);
+
+        truncatedBold = _bold;
+        for (uint256 i = 0; i < numBranches; ++i) {
+            if (branch[i].redeemable && branch[i].proportion > 0) {
+                uint256 extrapolatedBold = branch[i].redeemedBold * totalProportions / branch[i].proportion;
+                if (extrapolatedBold < truncatedBold) truncatedBold = extrapolatedBold;
```
bingen: I’m having hard time understanding this part. It seems that `truncatedBold` can be overridden at any branch iteration. Why do we know the last one is the good one? I’ll get back to it later, but maybe a comment may help.

bingen (commented):
It looks good in general. I still need to finish the review of `truncateRedemption`, but the important part in my opinion is that the wrapper `redeemCollateral` is fine. The frontend may show the amounts passed as collateral and their values in `$`, so even if one of the other helpers made a mistake the user would realize that is getting a bad deal.

contracts/src/RedemptionHelper.sol
```diff
@@ -0,0 +1,165 @@
...
+        uint256 _maxIterationsPerCollateral,
+        uint256 _maxFeePct,
+        uint256[] memory _minCollRedeemed
+    ) external {
+        require(_minCollRedeemed.length == numBranches, "Wrong _minCollRedeemed length");
+
+        RedemptionContext[] memory branch = new RedemptionContext[](numBranches);
+
+        for (uint256 i = 0; i < numBranches; ++i) {
+            branch[i].collToken = collateralRegistry.getToken(i);
```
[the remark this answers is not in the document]
danielattilasimon: I considered this too. The 3 calls to `collateralRegistry.getToken()` total about 1.35K gas, so that's the upper limit on how much we could save by switching to immutables. It makes the code a bit nasty though.

Or we could use storage, but unless we pass an access list with the TX, it would actually end up costing more, since cold slots take 2.1K to SLOAD.

In the end, I didn't think the 1.35K potential savings warranted adding the extra immutable optimization.

PR #1195, opened 2026-01-21
[Subgraph] Handle initiative unregistration

### LINKED ISSUE #1064: [Subgraph] Handle initiative unregistration
There are multiple places in the frontend where we make multicalls to fetch some state for each initiative. Over time, the number of batched calls will keep growing. At some point, we might need to filter this to initiatives that are either still active or the user still has some votes allocated towards. To that end, we should handle initiative unregistration in the subgraph and add a new field to the schema that lets us filter out unregistered initiatives.

cyril-dfi:


danielattilasimon (approved):
Looks good! I've started redeploying the subgraph, it'll probably take a few days to propagate.

PR #1196, opened 2026-02-02
[App] Users can override RPC and The Graph URLs

### LINKED ISSUE #1189: [App] Make it easy for user to fall back to a local frontend version (for emergencies)

cyril-dfi:
Fixes https://github.com/liquity/bold/issues/1189
Host the frontend on IPFS and let users use their own RPC / The Graph URLs
image

image


danielattilasimon (approved):
This is great!

Added some documentation for the new env var in `frontend/app/README.md`.

PR #1198, opened 2026-03-17
[App] Automatic interest rate

cyril-dfi:
image
image
image
image


cyril-dfi:
Won't add it to the codebase for now

